PR #96, opened 2016-08-05
fix(readme): rework guidelines and add limitation for images

tnguyen1:


README.md
```diff
@@ -1,37 +1,133 @@
...
+A livereload is also set up and allows any change to the `md` files to reload its html version in the browser.
+
 
 ## Contribute
 
+### Update existing pages
+
+In this case, a simple edit of the _.md_ file in the appropriate branch (one per version) is enough, the build process will do the rest.
+
+<u>Example</u>: To edit the **actors.html** page in the 7.3 version, checkout the **versions/7.3** branch and edit the `md/actors.md` file.
```
redboul: currently, there is no version per branch and we still did not clarify this...
The way it is right now is that the `master` branch contains 7.3 and the `dev` branch contains 7.4.

README.md
```diff
@@ -1,37 +1,133 @@
...
+
+### Add new pages
+
+#### Create new pages
+
+Checkout the version you want this page to be available in and create the _.md_ file in the **md** folder. The **master** branch contains documentation applicable to the latest version of Bonita BPM.
```
redboul: same remark than above

README.md
tnguyen1: Agreed - should we create a `versions/7.3` branch now that 7.3 is released?
Or change to follow the development branch model with master/dev ?

redboul: A dev branch has been created but I do not think that this model can be applied here.
We should create a 7.3 branch and have the master branch pointing to the next version.
What do you think?

tnguyen1: OK with creating a 7.3 branch and master pointing to 7.4.
Then how to deal with changes applicable to both branches? Shall the author create 2 pull requests (one on 7.3 branch, one on master branch)? If so, we need to communicate the process I guess.

redboul: we should proceed the same way than with development project : rebase master branch on previous branch

tnguyen1: I'm not sure we can rebase master branch on previous version branches.
I think each version branch is likely to have its own lifecycle.

Basically we need to consider these cases (given 7.3 and 7.4=current version):
- update on 7.3 only
- update on 7.4 only
- update on both 7.3 and 7.4

Better discuss about it when you are available.

tnguyen1: Up / Are we clear on how we manager versions/branches?

PR #102, opened 2016-08-09
Fix Ubuntu install

lio-p:


md/ubuntu-openjdk-tomcat-postgresql.md
```diff
@@ -136,45 +138,72 @@ Warning: make sure you stop Tomcat before performing following operations: `sudo
...
-* Copy and overwrite `server.xml` file: `sudo cp ~/BonitaBPMSubscription-6.4.0-deploy/Tomcat-7.0.55/conf/server.xml /etc/tomcat7/server.xml`
+* Copy and overwrite `logging.properties` file: `sudo cp ~/BonitaBPMSubscription-x.y.z-deploy/Tomcat-7.0.67/conf/logging.properties /etc/tomcat7/logging.properties`
+* Copy and overwrite `context.xml` file: `sudo cp ~/BonitaBPMSubscription-x.y.z-deploy/Tomcat-7.0.67/conf/context.xml /etc/tomcat7/context.xml`
+* Copy and overwrite `server.xml` file: `sudo cp ~/BonitaBPMSubscription-x.y.z-deploy/Tomcat-7.0.67/conf/server.xml /etc/tomcat7/server.xml`
 * Edit `server.xml` (`sudo nano /etc/tomcat7/server.xml`) and comment out h2 listener line
 * Fix ownership on the copied files: `sudo chown -R root:tomcat7 /etc/tomcat7`
+* Copy the folder `platform_setup`: `sudo cp -R ~/BonitaBPMSubscription-x.y.z-deploy/platform_setup /opt/bonita/` and rename it: `sudo mv /opt/bonita/platform_setup /opt/bonita/setup`
+* Fix ownership on the copied files: `sudo chown -R tomcat7:tomcat7 /opt/bonita/setup`
+
+### Fix service tomcat7 issue
```
redboul: what is the issue actually?

lio-p: I can't tell actually. I spent a fair amount of time trying to figure a better way without success. The issue with Tomcat 7 in service mode is that it doesn't seem to execute the setenv.sh file, so the environment variables are not correctly set. 
Maybe it's related to my Linux distrib, or the Tomcat version, but that's the only workaround I found.

redboul:
Can you rebase the branch please?

md/ubuntu-openjdk-tomcat-postgresql.md
redboul: ok, it may be linked to your distro because on my machine (based on a ubuntu 14.04, those lines are already present...

lio-p: Got it. So should we just leave it with a message like, with some linux distributions the setenv script is not executed correctly, in that case here is a workaround?

redboul: I am not sure we should add documentation for unsupported environment...
need to chek with @dtranquang

redboul: @lio-p can you tell us onto which distro you had the issue?
apparently, we should not add this troubleshooting fix for a specific distro...

redboul:
can you reopen this PR onto 7.3 branch please?
master branch has been deleted in favor of a more explicit 7.3 branch.

Thx

PR #120, opened 2016-09-06
fix(api-extensions): remove references to BONITA_HOME

redboul:
Closes [BS-15735](https://bonitasoft.atlassian.net/browse/BS-15735)

### DIFF
--- a/md/api-extensions.md
+++ b/md/api-extensions.md
@@ -52,8 +52,10 @@ A REST API extension example resource and API extension viewer page are provided
 
 To use the examples, define a [custom permissions](rest-api-authorization.md#custom-permissions-mapping) mapping for `demoPermission`:
 
-* Edit `$BONITA_HOME/client/tenants/1/conf/custom-permissions-mapping.properties`.
+* Retrieve platform configuration using the [setup tool](BonitaBPM_platform_setup.md#update_platform_conf).
+* Edit `platform_conf/current/tenants/<TENANT_ID>/tenant_portal/custom-permissions-mapping.properties`.
 * Add the following line: `profile|User=[demoPermission]`.
+* Push the changed platform configuration into database using the _setup tool_.
 * Restart Tomcat bundle
 
 


md/api-extensions.md
```diff
@@ -52,8 +52,10 @@ A REST API extension example resource and API extension viewer page are provided
 
 To use the examples, define a [custom permissions](rest-api-authorization.md#custom-permissions-mapping) mapping for `demoPermission`:
 
-* Edit `$BONITA_HOME/client/tenants/1/conf/custom-permissions-mapping.properties`.
+* Retrieve platform configuration using the [setup tool](BonitaBPM_platform_setup.md#update-platform-conf).
```
educhastenier: hash is incorrect: _update_platform_conf_

md/api-extensions.md
```diff
@@ -52,8 +52,10 @@ A REST API extension example resource and API extension viewer page are provided
...
+* Edit `$PLATFORM_CONF/current/tenants/1/tenant_portal/custom-permissions-mapping.properties`.
```
educhastenier: Did we explain somewhere **$PLATFORM_CONF** ?

educhastenier: Maybe replace tenant 1 by `<your_tenant_id>`

redboul: it is used elsewhere but I do not know if we have

educhastenier: Even if used elsewhere, I think we should never need to navigate to another page to understand the meaning of a variable. A doc page should be self-sufficient in most cases.

educhastenier: Ok, after comparing with other pages `platform_conf/current/tenants/1/...` is enough.

educhastenier:
Commentaire integration slack - github

md/api-extensions.md
redboul: so you think we need to explain this $PLATFORM_CONF here?

educhastenier: no, just change `$PLATFORM_CONF` by `platform_conf` and that's enough

tnguyen1:
retest this please

PR #123, opened 2016-09-06
fix(actors): use direct link to 6.5 documentation page on anonymous user

redboul:
Using relative links from current document to legacy one do not work
because of link management during md to html transformation.
Use absolute link instead.

Closes [BS-15273](https://bonitasoft.atlassian.net/browse/BS-15273)


educhastenier:
retest this please

educhastenier:
Works fine.

PR #127, opened 2016-09-07
fix(tenant): warning on theme loading on new tenant

redboul:
Closes [BS-15700](https://bonitasoft.atlassian.net/browse/BS-15700)


cpuy:
LGTM. Could this be merged ?

educhastenier:
What is said here still needs to be confirmed, because Julien tested it on a biased environment

abirembaut:
I don't know if this has been fixed since, but in 7.3.0, what Julien says is true. I can confirm it.

PR #130, opened 2016-09-12
fix(doc): fix issue with table rendering

amottier:
fix(doc): fix issue with table rendering

In getting started page, table that list data model was not displayed properly: content column shift on the right.
Remove leading spaces characters should fix the issue.

### DIFF
--- a/md/getting-started-tutorial.md
+++ b/md/getting-started-tutorial.md
@@ -175,16 +175,16 @@ To define the business data model:
 3. Click on the temporary name, and change it to _TravelRequest_.
 4. Select the _TravelRequest_ business object type, and add its attributes. This table shows the attributes and their types:
 
-  | Variable  | Data type | Mandatory |
-  | --------- | --------- | -------- |
-  | userId  | Long  | yes  |
-  | departureDate  | Date  | yes  |
-  | numberOfNights  | Integer  | yes  |
-  | hotelNeeded  | Boolean  | yes  |
-  | destination  | String  | yes  |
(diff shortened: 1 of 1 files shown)


md/getting-started-tutorial.md
```diff
@@ -175,16 +175,16 @@ To define the business data model:
...
+| Variable  | Data type | Mandatory |
+| --------- | --------- | -------- |
+| userId  | Long  | yes  |
+| departureDate  | Date  | yes  |
+| numberOfNights  | Integer  | yes  |
+| hotelNeeded  | Boolean  | yes  |
+| destination  | String  | yes  |
+| reason  | String  | yes  |
+| status  | String  | yes  |
+| refusalReason  | String  | no  |
```
redboul: Is it normal that the table is not aligned with the ordered list ?

amottier: The extra two spaces does not align the table with the ordered list items but lead to an issue in columns content display on our web site. Easy fix is to remove the two spaces.

amottier:
Is there anything that still prevent merge of this modification? [Documentation in production on the website is not usable](http://documentation.bonitasoft.com/?page=getting-started-tutorial#toc1).

md/getting-started-tutorial.md
redboul: Actually, you need three spaces to be aligned with the content above. 
Can you change this plz?
there may be some other tables and text that can be aligned too.

PR #132, opened 2016-09-14
fix(wildfly) fix bonita-platform-init-community-custom.properties pat…

cpuy:
…h in wildfly bundle


redboul:
This applies on jboss bundle of 7.3 too, can you open the PR on 7.3 instead plz ?

cpuy:
Yep

cpuy:
Will it be properly rebased ?

cpuy:
To be reopened on 7.3

PR #134, opened 2016-09-14
fix(wildfly) fix bonita-platform-init-community-custom.properties pat…

cpuy:
…h in jboss bundle


educhastenier (approved):
Good job, Colin.

PR #136, opened 2016-09-14
add Jenkinsfile to generate html from md

tnguyen1:
Adding Jenkinsfile to use Jenkins Pipeline as code with multiple branches.
@see https://jenkins.io/blog/2015/12/03/pipeline-as-code-with-multibranch-workflows-in-jenkins/

This will automatically create one job per version.
@see http://doc.rd.lan/job/doc/

Basically this will add http://doc.rd.lan/job/doc/branch/7.3 and http://doc.rd.lan/job/doc/branch/7.4 jobs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,41 @@
+node {
+   stage 'Checkout'
+   git branch: '7.3', url: 'git@github.com:bonitasoft/bonita-doc.git'
+   
+   stage 'Build'
+   // Get NodeJS custom tool
+   def nodejsHome = tool name: 'NodeJS_424', type: 'com.cloudbees.jenkins.plugins.customtools.CustomTool'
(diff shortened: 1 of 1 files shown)


redboul:
Is it mandatory to have this file on root folder?

tnguyen1:
Apparently yes as per documentation:

> Just having this file in the source code repo root would mean that -
> 
> Jenkins will automatically recognize this branch and create appropriate jobs by itself.
> Quick, 1-step code checkout using: “checkout scm” in your workflow
> Every time a new change is pushed to this branch, the branch is built and the commit status gets updated.
> When the branch is destroyed in the repository, or if Jenkinsfile is removed, the corresponding job gets destroyed from Jenkins automatically (You can retain these jobs and/or archive the builds for audit/compliance requirements using the retention property - Orphan Item strategy)

What kind of issue are you thinking of?

PR #143, opened 2016-09-22
fix(ci): use 'checkout scm' instead of checking out specific branch +…

tnguyen1:
… automatically trigger 'push' jobs


tnguyen1:
retest this please

PR #146, opened 2016-09-27
fix(doc): add warning build not supported on Windows

amottier:
Due to BS-8375 we don't support compilation of Bonita BPM on Windows.
Add a warning in documentation to avoid bug reports from users.

### DIFF
--- a/md/building-community-edition-from-source.md
+++ b/md/building-community-edition-from-source.md
@@ -15,7 +15,14 @@ To build the product, you need the following:
 * NodeJS
 
 ::: info
-We provide build script for Linux only. Therefore it is recommended to use Linux. If you don't have Linux you can either use a virtual machine or port the build script to Mac or Windows.
+We provide build script for Linux only. Therefore it is recommended to use Linux. If you don't have Linux you can either use a virtual machine or port the build script to Mac.
+:::
+
+<!---
+BS-8375
+--->
+::: warning
+Note that you can run Bonita BPM on Windows but you cannot build Bonita BPM on Windows (known limitation).
 :::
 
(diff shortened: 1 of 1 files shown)


educhastenier:
Why cannot we build Bonita BPM on Windows? What is the limitation?

amottier:
According to [BS-8375](https://bonitasoft.atlassian.net/browse/BS-8375) engine tests cannot be run on Windows.

educhastenier:
To build the product, you don't require to run the tests

amottier:
Agree but usually if you build the software it's because you change it and so it's likely that you want to run tests to make sure you didn't break anything.

cpuy:
Might be a little bit strong to say that Bonita BPM is not buildable on windows since only Engine test don't pass no windows ?

@educhastenier Would it be a proper solution to make engine tests pass on windows ?

educhastenier:
Don't know how much work it is to make tests pass on Windows.
Do we know what the problem is?

md/building-community-edition-from-source.md
```diff
@@ -15,7 +15,14 @@ To build the product, you need the following:
 * NodeJS
 
 ::: info
-We provide build script for Linux only. Therefore it is recommended to use Linux. If you don't have Linux you can either use a virtual machine or port the build script to Mac or Windows.
+We provide build script for Linux only. Therefore it is recommended to use Linux. If you don't have Linux you can either use a virtual machine or port the build script to Mac.
+:::
+
+<!---
+BS-8375
```
educhastenier, line 22: Is it a good practice to reference JIRA tickets in commentary?

PR #150, opened 2016-09-29
fix(uid) remove ui-designer log informations

cpuy:
Since UI Designer log is not created in tomcat_home/logs, it is preferable to remove information about this log until we find a proper way to deal with it

Closes [BS-15813](https://bonitasoft.atlassian.net/browse/BS-15813)

### DIFF
--- a/md/log-files.md
+++ b/md/log-files.md
@@ -4,7 +4,7 @@
 **Note:** For Performance, Efficiency, and Teamwork editions only.
 :::
 
-During development, the Studio log file contain messages about development activities, including any errors.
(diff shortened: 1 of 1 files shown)


md/log-files.md
```diff
@@ -4,11 +4,11 @@
 **Note:** For Performance, Efficiency, and Teamwork editions only.
 :::
 
-During development, the Studio log file and UI designer log file contain messages about development activities, including any errors. 
+During development, the Studio log file contain messages about development activities, including any errors.
```
redboul: typo: contains

PR #158, opened 2016-10-03
AFTER security-domains opening tag instead of before

marielle-spiteri:
The insertion should be done AFTER and not before the <security-domains> tag in sentence: 

"4 - Edit JBOSS_HOME/standalone/configuration/standalone.xml and add the BonitaAuthentication module. before the <security-domains> tag, insert these lines (specifying the relevant IP addresses and port numbers)"


tnguyen1:
add to whitelist

PR #165, opened 2016-10-06
fix(howtos): add How to in page titles

redboul:


md/uid-case-overview-tutorial.md
```diff
@@ -1,4 +1,4 @@
-# Create a case overview using UI Designer
+# How to create a case overview using UI Designer
```
educhastenier: using **the** UI Designer

PR #176, opened 2016-10-18
BS-15942 Order is not working in the REST API

pierre-yves-monnet:
Add a description in `/API/bpm/humanTask` to give explicitly the list of attribute to order the list (because this is not the "any attribute in the result can be ordered". See BS-15942 and the explanation of Anthony.

### DIFF
--- a/md/bpm-api.md
+++ b/md/bpm-api.md
@@ -481,6 +481,19 @@ Retrieve humanTask objects that match the specified filters.
   `GET`
 * **Data Params**  
   [Standard search parameters](rest-api-overview.md#resource_search) are available.  
+  
+  * o: only parameters available in HumanTaskInstanceSearchDescriptor.java and in the result are accepted: 
+	* assignee_id
+	* diplayname
+	* dueDate
+	* name
+	* priority
+	* processid
+	* id
+	* reached_state_date
+	* state
+	
+	
   You can also use the [flow node data params](#flownode-search-data-params).  
   For instance, you can filter on:
   * `assigned_id={user_id}`: retrieve only the human tasks assigned to the specified ID. For example, retrieve the human tasks assigned to user with id 2: `/API/bpm/humanTask?p=0&c=10&f=assigned\_id%3d2`
--- a/md/bpm-api.md
+++ b/md/bpm-api.md
@@ -482,16 +482,16 @@ Retrieve humanTask objects that match the specified filters.
 * **Data Params**  
   [Standard search parameters](rest-api-overview.md#resource_search) are available.  
   
-  * o: only parameters available in HumanTaskInstanceSearchDescriptor.java and in the result are accepted: 
-	* assignee_id
-	* diplayname
-	* dueDate
-	* name
-	* priority
-	* processid
-	* id
-	* reached_state_date
-	* state
+  * o: only fields available in [HumanTaskInstanceSearchDescriptor] (http://documentation.bonitasoft.com/javadoc/api/7.3/org/bonitasoft/engine/bpm/flownode/HumanTaskInstanceSearchDescriptor.html) and in the result are accepted: 
+    * case_id
+    * root_case_id
+    * parent_case_id
+    * process_id
+    * state
+    * type    	
+    * assignee_user_id
+    * priority
+
 	
 	
   You can also use the [flow node data params](#flownode-search-data-params).  


md/bpm-api.md
```diff
@@ -481,6 +481,19 @@ Retrieve humanTask objects that match the specified filters.
   `GET`
 * **Data Params**  
   [Standard search parameters](rest-api-overview.md#resource_search) are available.  
+  
+  * o: only parameters available in HumanTaskInstanceSearchDescriptor.java and in the result are accepted:
```
rbioteau: Use
 `[HumanTaskInstanceSearchDescriptor](http://documentation.bonitasoft.com/javadoc/api/7.3/org/bonitasoft/engine/bpm/flownode/HumanTaskInstanceSearchDescriptor.html)`
instead of HumanTaskInstanceSearchDescriptor.java

rbioteau: only _fields_ available in ...

redboul:
like I said [here](https://github.com/bonitasoft/bonita-doc/pull/175), list is wrong due to handling of parameter by  REST API `Datastore` class

md/bpm-api.md
pierre-yves-monnet: Done / Done (is that possible to click on a OK somewhere ? )

pierre-yves-monnet:
This is not very simple for us which not have the source in Eclipse to get the value...
In :
https://github.com/bonitasoft/bonita-web/blob/master/server/src/main/java/org/bonitasoft/web/rest/server/datastore/bpm/flownode/AbstractHumanTaskDatastore.java#L76

I see the list : (no comment in the code :-( but I assume this is all HumanTaskItem.xxx like HumanTaskItem.ATTRIBUTE_CASE_ID but where is the VALUE of the constante ? According the state of the art, it should be case_id an so one..
I look in https://github.com/bonitasoft/bonita-web/blob/master/common/src/main/java/org/bonitasoft/web/rest/model/bpm/flownode/HumanTaskItem.java but there are no definition of constante here ( ?? ) which is strange : constant should be defined in the class, so what is the logic here to define a constante ? 
So I update the information based on the "state of the art".

pierre-yves-monnet:
I push again: is change are visible in the pull request of do I need to do something else ?

md/bpm-api.md
```diff
@@ -481,6 +481,19 @@ Retrieve humanTask objects that match the specified filters.
   `GET`
 * **Data Params**  
   [Standard search parameters](rest-api-overview.md#resource_search) are available.  
+  
+  * o: only fields available in [HumanTaskInstanceSearchDescriptor] (http://documentation.bonitasoft.com/javadoc/api/7.3/org/bonitasoft/engine/bpm/flownode/HumanTaskInstanceSearchDescriptor.html) and in the result are accepted:
```
redboul, line 485: I still disagree on the fact to talk about the engine class because it the Web class that manages parameters first.

md/bpm-api.md
```diff
@@ -481,6 +481,19 @@ Retrieve humanTask objects that match the specified filters.
   `GET`
 * **Data Params**  
   [Standard search parameters](rest-api-overview.md#resource_search) are available.  
+  
+  * o: only fields available in [HumanTaskInstanceSearchDescriptor] (http://documentation.bonitasoft.com/javadoc/api/7.3/org/bonitasoft/engine/bpm/flownode/HumanTaskInstanceSearchDescriptor.html) and in the result are accepted: 
+    * case_id
```
redboul, line 486: this is not the actual parameter name, they are, generally, in camel case

redboul (changes requested):
There is a big class hierarchy that is very hard to read on the FlowNodeItem and its subclasses (and don't even talk about [IFlowNodeItem](https://github.com/bonitasoft/bonita-web/blob/master/common/src/main/java/org/bonitasoft/web/rest/model/bpm/flownode/IFlowNodeItem.java) which holds some of the values you are looking for). We know it is a mess. That is why it is so badly documented and it is difficult to maintain it...

md/bpm-api.md
cpuy: Since it is a rest api doc. We don't want to reference any Java class. We just want to list available options here

nchabanoles:
The REST API is managing its own list of attributes that can be used. Referencing the Engine class is not correct. Another PR should be created to provide the list of attributes that are allowed.

PR #178, opened 2016-10-19
fix(platform-api): replace /API/platform/licence by /API/platform/lic…

JeremJR:
…ense


educhastenier:
No need to duplicate the PR on 7.4 as it will be rebase on 7.3

PR #184, opened 2016-10-26
fix(tuto case overview) add info on URL parameter creation

julienmege:
set the URL parameter name


md/uid-case-overview-tutorial.md
```diff
@@ -14,6 +14,7 @@ So just add a new variable with:
 ```css
```
redboul: why is it CSS?

redboul (commented):
LGTM without this css stuff that may have been removed previously...

PR #192, opened 2016-11-10
Update bpm-api.md

HappyBonita:
Add limitation on "Get a case variable" to avoid problems

### DIFF
--- a/md/bpm-api.md
+++ b/md/bpm-api.md
@@ -1850,7 +1850,7 @@ The methods used for this resource are:
       "type":"java.lang.Float"
     }
     ```
-::: warning _Limitations_  It is not compatible to render no primitif Java variable :::
+::: warning **Limitations** When using a Java Object variable,  `value` field contains the `toString()` value of the object, you may implement a `toString()` on your Java Object that serialized the object into a json string ( eg:  [Jackson ObjectMapper](http://wiki.fasterxml.com/JacksonInFiveMinutes) ) :::  
   
 
 #### Update a case variable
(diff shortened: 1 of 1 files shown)


md/bpm-api.md
```diff
@@ -1850,6 +1850,9 @@ The methods used for this resource are:
       "type":"java.lang.Float"
     }
     ```
+* **Limitations**
+  It is not compatible to render Complexe Data Type variable
```
rbioteau: What is a Complex~~e~~ datatype variable ?

HappyBonita:
or any custom class.

md/bpm-api.md
educhastenier: He just meant **complex data type** ~~variable~~

rbioteau:
Data type create this way just generate Java pojo
The variable has not a Complex data type it has a Java data type.
Are you sure that java variable are not supported here ?

md/bpm-api.md
rbioteau: You may also use 

```
::: warning _Limitations_  It is not compatible to render Complexe Data Type variable :::
```

To have a pretty warning display in doc.

rbioteau:
I just tried.
It 'works' **but** value contains the the toString value of the java object.
It means that in order to make it work _properly_ you have to implement the toString method of your pojo.
eg: using jackson lib to serialize the object as a JSON string.

IMHO Complex Data type has nothing to do with that and should note be mentionned here.

PR #205, opened 2016-11-18
fix(cluster): convert single node to cluster

laurentleseigneur:
* document how to update quartz sheduler name

Closes [BS-15661](https://bonitasoft.atlassian.net/browse/BS-15661)


baptistemesta:
LGTM :+1:

PR #206, opened 2016-11-18
remove config for business data

NathalieC:


baptistemesta:
some description on why we remove that?
+ please next time follow https://github.com/bonitasoft/bonita-developer-resources/blob/master/CONTRIBUTING.MD#-commit-format
I will update the commit name/description on merge

PR #207, opened 2016-11-21
Update translation guidelines

amottier:
fix(doc): update to follow new translation contrib process

Reflect the latest change on Crowdin public project in the documentation.


md/languages.md
```diff
@@ -15,7 +15,9 @@ Officially supported language means that:
 
 ## Select BPM Portal user interface language
 
-For the web desktop version: go to the top right corner, click on _**Settings**_/_**Language**_ and select a language from the list.
+For the web desktop version:
```
educhastenier: What do you mean `For the web desktop version` ?

md/languages.md
```diff
@@ -25,23 +27,20 @@ For the web mobile version:
...
 
-1. Go to [Bonita BPM translation project](http://translate.bonitasoft.org/)
-1. Click the **_Download_** icon to download the zip archive file containing the translations into all the available languages. The archive is called `bonita-bpm-60.zip` even for Bonita BPM 7 versions.
-1. Go to the `master/bonita-web/portal` folder.
-1. Find and extract the `.po` files for the language you want to add (several files for a single language). Each `.po` file has a language indicator and a locale indicator. For example, the files for the Brazilian Portuguese language end with `pt_BR.po`.
-1. For Subscription editions, you also need to get the files from `master/bonita-web-sp/portal` folder.
-1. Copy the `.po` files to a new folder of your choice, preferably in a sub-folder inside your bundle. E.g. `[TOMCAT_INSTALL_FOLDER]/i18n/`. For Bonita BPM Studio, Tomcat installation folder is located in `/workspace/tomcat`.
+1. Go to [Bonita BPM translation project](http://translate.bonitasoft.org/).
+1. Select the language you are interested in.
+1. Browse the file tree to `7.3.x/bonita-web/portal` folder.
```
cpuy: Does that mean that this page should be updated for each bonita version ?
Would it be preferable to add a section explaining the version (correspond to your current bonita blabla) and then use a variable or a generic thing for the version ?

md/languages.md
```diff
@@ -25,23 +27,20 @@ For the web mobile version:
...
+1. For each file in the folder: open it and in the **_File_** menu click on **_Download_**. Each `.po` file has a language indicator and a locale indicator. For example, the files for the Brazilian Portuguese language end with `pt_BR.po`.
+1. For Subscription editions, you also need to get the files from `7.3.x/bonita-web-sp/portal` folder.
```
cpuy: Same comment than above

cpuy:
Very nice, way more clear and also more robust since the user don't need to download the whole zip and therefore has always the latest translations.

See small notes above and also I guess a rebase is needed to remove extra commits that do not need to be there

md/languages.md
cpuy: I think here we should speak about the 'toggle hidden files' button if needed since when a translation is completed for a language, files are hidden

laurentleseigneur: "7.3.x" may be replaced by ${varVersion} as explained here: https://github.com/bonitasoft/bonita-doc/blob/7.3/CONTRIBUTING.md since it applies to 7.3+ versions

cpuy: Awesome !!! Glad to have someone that know the doc tooling.
Here is a more [direct link](https://github.com/bonitasoft/bonita-doc/blob/7.3/CONTRIBUTING.md#variables)

So I guess it should be  `${varVersion}.x`

laurentleseigneur: @cpuy good point, ${varVersion} is supposed to be replaced by current documentation version, that matches product minor version but excludes `.maintenanceVersion` e.g. in 7.3.4, ".4" will be ignored  so `${varVersion}.x` seems to be the right option

amottier: I use `${varVersion}.x` as suggested.
I also include indication about 'toggle hidden files' button.

md/languages.md
amottier: Also updated.

md/languages.md
```diff
@@ -15,33 +15,33 @@ Officially supported language means that:
...
+1. Select a new language from the drop-down list
 
 ## Add a language to Bonita BPM Portal
 
-::: warning
-We are currently updating the translation process. Among other things this update impacts the Crowdin project structure. Instruction below might be temporarily inaccurate.
-:::
-
 ::: info
 Instructions below assume that the language you want to add is already available in the community translation project. If translation is not available see [Translate Bonita BPM Portal](#Translate-Bonita-BPM-Portal) for instruction about how to collaborate to a new or ongoing translation.
```
cpuy: link is not working

md/languages.md
```diff
@@ -15,33 +15,33 @@ Officially supported language means that:
...
-1. Click the **_Download_** icon to download the zip archive file containing the translations into all the available languages. The archive is called `bonita-bpm-60.zip` even for Bonita BPM 7 versions.
-1. Go to the `master/bonita-web/portal` folder.
-1. Find and extract the `.po` files for the language you want to add (several files for a single language). Each `.po` file has a language indicator and a locale indicator. For example, the files for the Brazilian Portuguese language end with `pt_BR.po`.
-1. For Subscription editions, you also need to get the files from `master/bonita-web-sp/portal` folder.
-1. Copy the `.po` files to a new folder of your choice, preferably in a sub-folder inside your bundle. E.g. `[TOMCAT_INSTALL_FOLDER]/i18n/`. For Bonita BPM Studio, Tomcat installation folder is located in `/workspace/tomcat`.
+1. Go to [Bonita BPM translation project](http://translate.bonitasoft.org/).
+1. Select the language you are interested in.
+1. Make sur you click on the "Toggle Hidden Files" button ![Toggle hidden files button](images/crowdin_toggle_hidden_files.png) to see already fully translated files.
```
cpuy: Make sure ?

md/languages.md
cpuy: should it be `[Bonita BPM translation project](http://translate.bonitasoft.org/` ?

cpuy:
Nice.
Please see 2 small notes above
Deployed on http://doc.rd.lan:50207/7.3?page=languages

md/languages.md
amottier: No I try to insert a link to an anchor but forget to declare the anchor (I use GitHub syntax instead and doesn't seems to work).

md/languages.md
amottier: Fixed, thanks.

md/languages.md
cpuy: oh yes... 
you need to follow this https://github.com/bonitasoft/bonita-doc/blob/7.3/CONTRIBUTING.md#anchors to achieve what you want

PR #222, opened 2017-01-27
Updated Getting Started Tutorial to add changing target URL to an exp…

superbeagle:
I updated the tutorial near the end and added changing the Target URL to an expression rather than constant. It was not clear and led to errors when trying to run the example.


superbeagle:
definitely

PR #225, opened 2017-02-02
update import doc, details on how to manage conflicts with the new tool

alachambre:


md/import-and-export-a-process.md
```diff
@@ -20,6 +27,23 @@ You can import diagrams that were exported from the following tools:
 
 You can also [migrate a process diagram from Bonita Open Solution 5.x](migrate-a-process-from-bonita-open-solution-5-x.md).
 
+### How to manage conflicts?
```
rbioteau: Do we really want to document this ? I don't think so

PR #227, opened 2017-02-03
7.4.2 bug fixes list

dtranquang:


educhastenier:
I realize the titles are not very explicit, I would be great to reformulate them.

educhastenier:
Are you sure we did not forget any bug in the list? I will not double-check that.

PR #237, opened 2017-02-17
Update event-handlers.md

marielle-spiteri:
Added: ::: warning
In a cluster environment, Event handlers must be Serializable (implements Serializable and have all its fields serializable or transient) because they are shared with other nodes of the cluster.
:::


educhastenier:
retest this please

PR #244, opened 2017-03-02
doc(Weblogic): remove support for weblogic

educhastenier:
Closes [BS-16256](https://bonitasoft.atlassian.net/browse/BS-16256)


md/release-notes.md
```diff
@@ -7,3 +14,15 @@
  
  Only in the case you have created custom widgets using this silent library, you need to update them by adding UI Bootstrap js as an asset.
 
+
+### Full Java 8 support
+Bonita BPM 7.5.0+ is now fully Java 8 compliant.  
+Java 8 syntax and specific APIs can now be fully used in Connectors, Actor filters, Groovy Scripts, Custom REST APIs, Custom Event Handlers ...
```
nchabanoles: We should not have "etc" in this sentence as it does not help the user to understand wether he/she can use Java 8 when implementing an extension point.
Either list all the extension point or state that all Extension Points can now be implemented in Java 8 (This will avoid to have to maintain the list accurate).
Can we link to the page of stable extension points?

educhastenier: Is there such a page?

educhastenier: Found it, I'll do that.

md/release-notes.md
```diff
@@ -7,3 +14,15 @@
  
  Only in the case you have created custom widgets using this silent library, you need to update them by adding UI Bootstrap js as an asset.
 
+
+### Full Java 8 support
+Bonita BPM 7.5.0+ is now fully Java 8 compliant.  
+Java 8 syntax and specific APIs can now be fully used in Connectors, Actor filters, Groovy Scripts, Custom REST APIs, Custom Event Handlers ...  
+As a consequence, JRE or JDK 8+ is required to run. JRE / JDK 7 is no longer supported to run Bonita BPM 7.5.0+
```
nchabanoles: I will avoid the term supported here (to avoid confusion with the support activity).
I would say that Bonita BPM 7.5.0+ cannot be operated on Java 1.7 anymore.

Questions:
Aren't we suppose to use a variable instead of writing Bonita BPM? (For OEM for instance)
Can we have a Troubleshooting section with the error message you get if you use Java 7 instead of Java 8? So that this page can be found by a user when searching for the error.

md/release-notes.md
```diff
@@ -7,3 +14,15 @@
...
+
+### Removed support for Oracle Weblogic application server
+Bonita BPM 7.5.0+ does not support Oracle Weblogic application server anymore.  
+Refer to [full software requirements](hardware-and-software-requirements.md) for more details on supported environments.
+
+### Other updates
+* Hibernate is updated to 4.3.11, to support JPA 2.1 specification
```
nchabanoles: Hibernate has been updated from version XYZ to 4.3.11, to support JPA 2.1 specification

md/release-notes.md
educhastenier: Is it the purpose of the release-notes page to include a Troubleshooting section?

PR #249, opened 2017-03-08
doc(SAML): add a few details to make this tutorial more clear/explicit

educhastenier:


educhastenier:
retest this please

md/single-sign-on-with-saml.md
```diff
@@ -77,11 +81,13 @@ To configure Bonita BPM for SAML:
...
     ```
     
-    Make sure to [set the right tenant admin username](multi-tenancy-and-tenant-configuration#toc2). It is recommended to also replace the value of the passphrase (property auth.passphrase) which is used by the engine to verify the authentication request. The value must be the same as in the file bonita-tenant-sp-custom.properties.  
+    Make sure to [set the right tenant admin username](multi-tenancy-and-tenant-configuration#toc2).
+    It is recommended to also replace the value of the passphrase (property auth.passphrase) which is used by the engine to verify the authentication request.
+    The value must be the same as in the file **bonita-tenant-sp-custom.properties**.  
     If you need some users to be able to log in without having an account on the IDP, you can authorize it by setting the property `saml.auth.standard.allowed` to true. Users will then be able to log in using the portal login page (/loging.jsp) provided they have a bonita account and their password is different from their username.
 
 3. In the tenant_engine folder of each existing tenant: “$TOMCAT_HOME/setup/platform_conf/current/tenants/<TENANT_ID>/tenant_engine/
```
abirembaut: I guess we should use back quotes for this path too.

abirembaut (changes requested):
LGTM. Just a small note

PR #257, opened 2017-03-30
fix(rest-apis) set businessDataReference rest api deprecated

cpuy:
businessDataReference API is deprecated since 7.0.0 and we should use context API instead

Cloes [BS-15978](https://bonitasoft.atlassian.net/browse/BS-15978)


educhastenier:
Looks good to me :+1:

PR #260, opened 2017-04-04
Support PDF generation of a single page

lio-p:
Add a new parameter —file to the npm run pdf command that takes a page
name.

### DIFF
--- a/scripts/convert2PDF.js
+++ b/scripts/convert2PDF.js
@@ -24,12 +24,12 @@
        return;
      }
    }
-   let file = argv.file;
+  let file = argv.file;
 
   
   
   const readdirPromise = denodeify(fs.readdir);
-  let outfile = `Bonita-BPM-documentation`;
+  let outfile = (file ? file : `Bonita-BPM-documentation`);
 
   const flattenedTaxo = flattenTaxonomy(taxonomy);
   let htmlFile = '', count = 0;  
--- a/scripts/convert2PDF.js
+++ b/scripts/convert2PDF.js
@@ -26,28 +26,31 @@
    }
    let file = argv.file;
 
+  
+  
   const readdirPromise = denodeify(fs.readdir);
   let outfile = `Bonita-BPM-documentation`;
 
   const flattenedTaxo = flattenTaxonomy(taxonomy);
   let htmlFile = '', count = 0;  
-
-  if (!file) {
-    rx.Observable.from(flattenedTaxo.filter(fileName => fileName && !fileName.match(/^https?:\/\//)))
-    .select(html => {
-      return 'build/html/' + html + '.html'; 
-    }).bufferWithCount(3).toArray().subscribe(filesArray => {
-      convertLastFilesToPDF(filesArray.reverse());
-    });
-  } else {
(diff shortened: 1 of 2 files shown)


educhastenier:
Why didn't you base your Pull Request on 7.3?

lio-p:
I assumed we should work on the most recent active branch. Should we word on 7.3 branch?

amottier:
Actually modification that apply to all version should be done on the older version branch (i.e. 7.3). They are then replicated on other branch after merging is done on 7.3.

educhastenier:
Don't worry, the propagation of the change in taken care of by R&D.
See [EDITED]Internal documenation on PR mechanism[EDITED] for more details on the procedure.

scripts/convert2PDF.js
```diff
@@ -24,19 +24,31 @@
...
+      convertLastFilesToPDF(filesArray.reverse());
+    });
+  } else {
+    console.log("Generate PDF for file: " + file);
+    rx.Observable.from(flattenedTaxo.filter(fileName => fileName && !fileName.match(/^https?:\/\//) && fileName == file))
+    .select(html => {
+      return 'build/html/' + html + '.html'; 
+    }).bufferWithCount(3).toArray().subscribe(filesArray => {
+      convertLastFilesToPDF(filesArray.reverse());
+    });
```
cpuy: seems to have a lot of duplicated code here.
Would you mind removing some duplication by extracting some method ?
It seems that only the filter function changes isn't it ?

Maybe you can do something like that
```
rx.Observable.from(flattenedTaxo.filter(findAGoodNameForThat(file))
 .select(html => {
       return 'build/html/' + html + '.html'; 
  }).bufferWithCount(3).toArray().subscribe(filesArray => {
       convertLastFilesToPDF(filesArray.reverse());
  });

function findAGoodNameForThat(file) {
  return function filter(filename){
       // find a better way to do that
      return (file && filename == file) || fileName && !fileName.match(/^https?:\/\//) 
   };
}
```

lio-p:
Includes feedback from Colin

cpuy:
👍 
I don't want to be picky but... should [outfile](https://github.com/bonitasoft/bonita-doc/blob/7.3/scripts/convert2PDF.js#L29) take the filename given in argument when generating pdf for one file ?

Let me know if you want (and if you have time) to do this modification otherwise I'll merge it has it is.

Not related to this PR but build scripts are quite a mess. Indeed running `npm run pdf -- --file assets` does not work out of the box since `npm run build && npm run taxonomy` need to be launched first as a prerequisite. Not user friendly.

Furthermore, on macos I get an error at the end of the generation
```
error:  Error: Command failed: pdftk 'build/Bonita-BPM-documentation-0.pdf' cat output /var/folders/71/yljm1qkn5q9bbxbpv6nft3c40000gn/T/tmp-2684JaQVyxm0aW9d
/bin/sh: pdftk: command not found
```
but pdf generation seems to be ok

lio-p:
Done for the output filename. 

Yeah I agree with you on the build script, it could be better. 

For pdftk, you can install it from mac OS, there are procedures described on the web.

cpuy:
👍

PR #262, opened 2017-04-13
fix(bdm): bring back bdm database configuration page

cpuy:
Closes [BS-15843](https://bonitasoft.atlassian.net/browse/BS-15843)

### DIFF
--- a/md/database-configuration-for-business-data.md
+++ b/md/database-configuration-for-business-data.md
@@ -1,0 +1,147 @@
+# Database configuration for business data
+
+## Configuration overview
+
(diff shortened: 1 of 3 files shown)


educhastenier (changes requested):
We decided to merge pages **Database configuration** and **Business Data database configuration**. Please keep it like that, and add documentation directly to the resulting page.

cpuy:
Ok would you mind pointing out what need to be done so ?

cpuy:
Ok then I do not see how we can do that. Since the main issue is how to configure bundles datasources.
Anyway this PR is not relevant but I leave the branch up

PR #288, opened 2017-05-11
fix(search): fix word search based documentation

cpuy:
Closes [BS-15989](https://bonitasoft.atlassian.net/browse/BS-15989)


tbouffard:
Changes match what is required in the BS, so 👍

PR #289, opened 2017-05-12
Delete page specify-cpu-cores.md in 7.4

pooravbos:
Removing this page from the documentation as part of cleaning the document of all cpu core based licensing references

PR #295, opened 2017-05-17
doc(licensing): remove cpu license reference link 7.4 wildfly (#295)

pooravbos:
Remove all references to CPU licensing


educhastenier:
will report change into #289

PR #300, opened 2017-05-18
doc(date-time): create thinner ToC

NathalieC:
move "###" level titles into "####" level to let them out of ToC


abirembaut:
LGTM

PR #303, opened 2017-05-19
fix(timetracker): refactoring of the timetracker documentation

danila-m:
Closes [BS-16544](https://bonitasoft.atlassian.net/browse/BS-16544)

### DIFF
--- a/md/performance-tuning.md
+++ b/md/performance-tuning.md
@@ -406,13 +406,39 @@ Be sure to configure these appropriately.
 
 ## Connector time tracker
 
-It is now possible to track the duration of actions in a connector using a new time tracker. The tracker service tracks several connector lifecycle operations, and produces a CSV file.  
+It is now possible to track the duration of actions in a connector using a new time tracker. The tracker service tracks several connector lifecycle operations.  
 This service can impact performance so is disabled by default.  
-It is configured by editing `bonita-tenant-community-custom.properties` and `bonita-tenants-custom.xml`.  
-To activate connector time tracking: 
+It is configured by editing the following parameters in `bonita-tenant-community-custom.properties`.  
 
-1. Change the value of `startFlushThread` from `false` to `true`.
-2. Uncomment 'flushEventListeners' and 'activatedRecords' beans.
+```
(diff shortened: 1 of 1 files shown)


md/performance-tuning.md
```diff
@@ -406,13 +406,39 @@ Be sure to configure these appropriately.
...
+#bonita.tenant.timetracker.memory.maxSize=1000000
+```
+To activate connector time tracking:
+1. Uncomment all the previous lines except ```## Time tracker```.
+2. Change the value of `startTracking` from `false` to `true`.
+
+The other parameters can be left at their default value, left commented, or set to the desired value. What each of them does:
+
+1. `maxSize` maximum of records that will be saved by the time tracker before a flush.
+2. `flushIntervalInSeconds` the interval beetween two flushes on the timetracker thread.
```
nchabanoles: If flushIntervalInSeconds and maxSize are defined what happens? Is there a flush every flushIntervalInSecondsand AND everytime maxSize is reached before the interval complete?

Or those attributes should not be defined at the same time?

md/performance-tuning.md
```diff
@@ -406,13 +406,39 @@ Be sure to configure these appropriately.
...
+3. `csv.activateAtStart` wether to save the result of the timetracker into a csv file.
+4. `csv.folder` the folder where to save the csv file.
+5. `csv.separator` the separator character in the csv file
+6. `memory.activateAtStart` wether to save the result of the timetracker in memory.
+7. `memory.maxSize` maximum amount of records saved in memory
```
nchabanoles: What happens when the maxSize is reached? Older record are lost? Is there a flush on CSV?
Can we have csv.activeAtStart and memory.activeAtStart enable at the same time?

nchabanoles (approved):
Only few questions but current changes are already bringing more information than previsous version.

danila-m:
"Can we have csv.activeAtStart and memory.activeAtStart enable at the same time?" the answer is yes. I added more precision in the description of the maxsizes options.

nchabanoles:
LGTM

PR #309, opened 2017-05-29
feat(SAML): integrate user test feedback

abirembaut:
- add IdP section
- add keys and certificates format note
- add explanation on how to test the login (URL)
- add logging configuration for troubleshooting

Covers [BS-16586](https://bonitasoft.atlassian.net/browse/BS-16586)

### DIFF
--- a/md/single-sign-on-with-saml.md
+++ b/md/single-sign-on-with-saml.md
@@ -136,6 +136,10 @@ To configure Bonita BPM for SAML:
 5. If your Identity provider (IdP) requires requests to be signed, generate a private key.
 For example on linux, you can use the command ssh-keygen, then go to “cd ~/.ssh” to retrieve the key from the file id_rsa (more id_rsa, then copy the key).
 
+::: info
+**Note:** The expected format for Keys and certificates is PEM (with or without the comment header and footer).  
+:::
+
 6. In the tenant_portal folder of each existing tenant: `$TOMCAT_HOME/setup/platform_conf/current/tenants/<TENANT_ID>/tenant_portal`,  
     edit the file **keycloak-saml.xml** to setup Bonita webapp as a Service provider working with your IdP.  
     + The entityID is the Service Provider given to your bonita installation. You can change it if you want but you need to provide it to your IdP.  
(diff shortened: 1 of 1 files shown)


NathalieC:
After Train the Trainers, a need has been raised to add a few sentences that explain:
- why logout has not been implemented (different behavior depending on the IdP)
- how to hide the logout button in Bonita

PierrickVoulet:
[Here ](https://access.redhat.com/documentation/en-us/red_hat_single_sign-on/7.0/html/securing_applications_and_services_guide/saml_2) is a complete doc for the keycloak (edited by RedHat, same as the one found on Keycloak but more readable).

The idea is to provide a minimal content removing all not mandatory parts (tags/attributes) in the default XML that define default behavior/value discussed in the RedHat documentation. I already did the exercise for a customer, it worked on the last RC without any issue.

abirembaut:
@NathalieC 
- There is already a clear explanation of the procedure to hide the logout button.
- As for single logout, all IdP do not support global logout and  there are different Single logout mechanism. But the real reason why we do not support it is that we didn't identified it as a common use case we need to support. Most existing customers are satisfied with hiding the logout button. I understand that it is not suitable for public computers, though. We could investigate more on the topic if it becomes a use case we need to support natively. Currently, you can implement your authentication Manager extending the SAML one to override the logout method. Should we say that in the doc ?

@PierrickVouletBonitasoft 
Thanks. we will create a new jira issue to remove the default attributes that should never be modified

NathalieC:
The updated page can also point to a blog post from @lio-p, based on this page, that explains how to setup SAML with SalesForce. It can be found here (for now): https://docs.google.com/document/d/1Bn3p6gFFqV7AkIzjkm5eSYRtpVAyYVwIeQmJLhI13ms/edit
@lio-p : could you send us the final location when it's published?

abirembaut:
IMHO, this should be handled in another PR and we should merge this one ASAP to help people trying to setup SAML with 7.5.

abirembaut:
up !

abirembaut:
It's been 2 month. Too late !
I merge this as there is no new feedback and if we want to add more content, I suggest opening a new PR.

PR #311, opened 2017-06-01
fix(bpm-api): list available order for archivedCase api

cpuy:
Closes [BS-16505](https://bonitasoft.atlassian.net/browse/BS-16505)


md/rest-api-overview.md
```diff
@@ -161,7 +161,6 @@ For a GET method that retrieves more than one instance of a resource, you can sp
 * p: index of the page to display
 * c: maximum number of elements to retrieve
 * o: order of presentation of values in response: must be either `attributeName ASC` or `attributeName DESC`. The final order parameter value must be URL encoded.
```
educhastenier, line 163: Maybe you should precise that possible values for attributeName may be different for every API

cpuy: As mentionned in BS the idea is to stay evasive so it sounds enough to me.

PR #316, opened 2017-06-06
fix(contract type): add doc on supported LONG type in pool contracts

educhastenier:
Closes [BS-16691](https://bonitasoft.atlassian.net/browse/BS-16691)

### DIFF
--- a/md/contracts-and-contexts.md
+++ b/md/contracts-and-contexts.md
@@ -43,7 +43,9 @@ Supported input types:
 4. **decimal:** a decimal value (e.g. 5.128)
 5. **file:** a document
 6. **integer:** a numeric value (e.g. 4)
+7. **long:** a long value (e.g. 3 000 000 000). Can be used at pool-level only (not at task-level). In called processes, this is helpful to receive IDs from call activities.
 7. **text:** a string
+
 
 Check **_multiple_** to specify that the input is a list of its primary type (for example, a list of integers).
 
--- a/md/contracts-and-contexts.md
+++ b/md/contracts-and-contexts.md
@@ -46,8 +46,11 @@ Supported input types:
 7. **long:** a long value (e.g. 3 000 000 000). Can be used at pool-level only (not at task-level). In called processes, this is helpful to receive IDs from call activities.
 7. **text:** a string
 
-
 Check **_multiple_** to specify that the input is a list of its primary type (for example, a list of integers).
+
+::: info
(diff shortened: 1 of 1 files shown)


amottier:
I think it would be useful to clarify how we can send such value as JSON when using the REST API.
Especially for long value that are in the ]2^53 -1 ; 2^63-1] range. Can we use any JSON number? Does the limit only apply to JavaScript number?...

educhastenier:
Can anyone with Web knowledge on that matter complete this PR with Antoine's remarks?

lcustodio:
We just verified and it's fine to use `long` type for POST. 
Without getting too technical, the POST goes straight to the server and the values are casted to contract type in java which guarantees accuracy. (The post send the content as string)

Tests were made with: 
* 2^53-1 (9 007 199 254 740 992) - Max value javascript
* 2^63-1 (9 223 372 036 854 775 807) - Valid long, *invalid* javascript number.

lcustodio (approved):
:shipit:

md/contracts-and-contexts.md
```diff
@@ -43,10 +43,15 @@ Supported input types:
 4. **decimal:** a decimal value (e.g. 5.128)
 5. **file:** a document
 6. **integer:** a numeric value (e.g. 4)
+7. **long:** a long value (e.g. 3 000 000 000). Can be used at pool-level only (not at task-level). In called processes, this is helpful to receive IDs from call activities.
 7. **text:** a string
 
 Check **_multiple_** to specify that the input is a list of its primary type (for example, a list of integers).
 
+::: info
+**:fa-info-circle:**  The type `long` can be used in a POST with JSON without precision lost. However, it is important to keep in mind that while manipulating numbers in javascript the max `integer` is 2^53-1 which is a smaller subset of max `long` type (2^63-1).
```
educhastenier: subset of **Java** max `long` type (2^63-1).

educhastenier:
@amottier happy with that?

amottier:
@educhastenier I think it would be useful to highlight in the note that it's possible to send the number as a JSON string if value is out of the JavaScript safe integer range.

Here is a sum up of two valid scenario:
If your value is in JavaScript safe integer range: Text widget (number) -> JavaScript number -> JSON number -> Contract java.lang.Long
If value is out of JavaScript safe integer range: Text widget (text) -> JavaScript String -> JSON String -> Contract java.lang.Long

educhastenier:
@amottier i suggest you contribute it directly than explain what to contribute, that would be quicker

educhastenier:
Looks good to me :+1:

PR #319, opened 2017-06-08
fix(rest-api): list order parameter for HumanTask API

cpuy:
Add availables values for `o` parameter for HumanTask API. Remove link between FlowNode API and HumanTask API parameters since those are different for the two APIs

Closes [BS-15942](https://bonitasoft.atlassian.net/browse/BS-15942)


abirembaut:
:+1:

PR #329, opened 2017-06-19
fix(mssql server): jdbc driver version clarification and troubleshooting

tbouffard:
In the 7.5.0+ version, user must use the sqlserver42.jar jdbc driver as Bonita runs with Java 8.
Also add a troubleshooting tip for the error that currently occurs with the Wildfly bundle.

Relates to [BS-16758](https://bonitasoft.atlassian.net/browse/BS-16758)

### DIFF
--- a/md/wildfly-bundle.md
+++ b/md/wildfly-bundle.md
@@ -216,3 +216,11 @@ For Oracle, rename it so that the name contains at least the word `oracle` or `o
 **Solution**: Ensure your running environment has a JDK or JRE 1.8 installed and set either JAVA or JAVA_HOME environment variable to point to it.
 
 ---
+
+**Issue**: When I start the Wildfly bundle configured to use a **Microsoft SQL Server** database, I get the error message `java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter`
+
+**Cause**: The WildFly configuration has not been properly updated (BS-16758)
+
+**Solution**: In the _<WILDFLY_HOME>/server/modules/com/sqlserver/main/module.xml_ file, add `<module name="javax.xml.bind.api"/>` in the list of dependencies
+
+---
--- a/md/wildfly-bundle.md
+++ b/md/wildfly-bundle.md
@@ -126,7 +126,7 @@ For production, you are recommended to use one of the supported databases, with 
 
 1. Make sure [your databases are created](database-configuration.md#database_creation) and [customized to work with Bonita BPM](database-configuration.md#specific_database_configuration).
 2. Edit file `<WILDFLY_HOME>/setup/database.properties` and modify the properties to suit your databases (Bonita BPM internal database & Business Data database)
-3. If you use **Microsoft SQL Server** or **Oracle** database, copy your database drivers in `<WILDFLY_HOME>/setup/lib` folder. 
+3. If you use **Microsoft SQL Server** or **Oracle** database, copy your [jdbc driver](database-configuration.md#proprietary_jdbc_drivers) in `<WILDFLY_HOME>/setup/lib` folder. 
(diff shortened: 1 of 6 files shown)


ugaston (approved):
Changes look good to me.

As a side note, adding some kind of reference to the required version of the driver here:
-> http://documentation.bonitasoft.com/?page=tomcat-bundle#toc5  (point 3.)
-> http://documentation.bonitasoft.com/?page=wildfly-bundle#toc5  (point 3.)
-> http://documentation.bonitasoft.com/?page=deploy-bundle#toc7  
 
would definitely improve the user's installation experience.

Thanks.

md/wildfly-bundle.md
```diff
@@ -216,3 +216,11 @@ For Oracle, rename it so that the name contains at least the word `oracle` or `o
 **Solution**: Ensure your running environment has a JDK or JRE 1.8 installed and set either JAVA or JAVA_HOME environment variable to point to it.
 
 ---
+
+**Issue**: When I start the Wildfly bundle configured to use a **Microsoft SQL Server** database, I get the error message `java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter`
+
+**Cause**: The WildFly configuration has not been properly updated (BS-16758)
```
educhastenier: BS-16758 does not mean anything for a non-internal bonitasofter. remove it.

educhastenier:
@ugaston what kind of reference would you like? This is just a change of version of the driver.
What do you suggest, more precisely?

md/wildfly-bundle.md
tbouffard: I wanted to point to an issue that will be documented in the release notes page (I hope in 7.5.2). If you think this is confusing, I remove it. @ugaston any feedbacks?

ugaston: Manu is probably right, here. I'm not sure it adds much value to the cause, knowing there is a bug associated..

ugaston:
@educhastenier As I said, it was more of a general comment (thus, not necessarily applicable to this very PR), and I meant that when one's following the documentation to install the product (whether Tomcat-, Wildfly-, or Deploy bundle), at one point, we are told to copy the corresponding JDBC driver into the application server's lib/ folder.

Well, say I am installing Bonita wih a MS SQL dabatase; I won't necessarily know the right version of the driver to use is the 6.0 (sqljdbc42.jar).... unless I start digging and diving into the various links of the documentation (out of my own initiative) and finally come across this mention of the JDBC driver version in the "Customize RDBMS to make it work with Bonita BPM" section of the "Database Configuration" chapter....

educhastenier:
@ugaston would it be ok if we add links that point directly to the sqlserver driver version?

ugaston:
@educhastenier Sure, that would definitely help.

By the way, if there were any similar limitation for Oracle DB drivers, the same would apply... (a link to the appropriate version would be welcome in the installation chapter).

Thanks.

md/wildfly-bundle.md
tbouffard: Ok, I am going to remove the ref to the associated bug

tbouffard:
Ok, as suggested by @ugaston, the documentation now explains how to retrieve the jdbc driver and the 3 bundle pages have a link to it.

Based on discussion with @nchabanoles and @laurentleseigneur, I also did that for the Oracle Database because we had the same need for this db.

ugaston (approved):
The new links (and the new "Proprietary Jdbc drivers" section) are great.

Thanks a lot!

laurentleseigneur:
LGTM
@tbouffard can you accept CLA

tbouffard:
@laurentleseigneur CLA signed

PR #343, opened 2017-07-17
feat(renaming): remove missing BPMs

danila-m:
Relates to [BS-16606](https://bonitasoft.atlassian.net/browse/BS-16606)

### DIFF
--- a/md/applications.md
+++ b/md/applications.md
@@ -44,7 +44,7 @@ To integrate a process into an application, you specify a link in an application
 ## Defining a new application
 
 ::: info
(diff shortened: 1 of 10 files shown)


md/applications.md
```diff
@@ -44,7 +44,7 @@ To integrate a process into an application, you specify a link in an application
 ## Defining a new application
 
 ::: info
-Since Bonita BPM 7.5.0, the definition of applications can be done from the studio, which makes it more straightforward to develop and package with other application artifacts.
+Since Bonita 7.5.0, the definition of applications can be done from the studio, which makes it more straightforward to develop and package with other application artifacts.
```
educhastenier: version 7.5.0 was still named Bonita BPM, so I think it should be kept as is...
to be discussed.

PR #347, opened 2017-07-18
fix(database): encoding and collation fix for PostgreSQL and SQL Server

amottier:
Current documentation requires to do something impossible:
- Configure SQL Server for UTF-8 (generic instruction for all database vendors)
- Configure PostgreSQL to be case insensitive (generic instruction for all database vendors). This is actually implemented by https://github.com/bonitasoft/bonita-engine/blob/master/services/bonita-persistence/bonita-persistence-hibernate/src/main/java/org/bonitasoft/engine/persistence/PostgresInterceptor.java

Close [BS-16880](https://bonitasoft.atlassian.net/browse/BS-16880)

### DIFF
--- a/md/database-configuration.md
+++ b/md/database-configuration.md
@@ -78,8 +78,6 @@ To create the database, we recommend you refer to your RDBMS documentation:
 * [SQL Server](https://technet.microsoft.com/en-us/library/dd207005(v=sql.110).aspx)
 * [MySQL](http://dev.mysql.com/doc/refman/5.5/en/database-use.html)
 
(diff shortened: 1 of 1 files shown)


tbouffard (changes requested):
When I talk about collations with @amottier, choosing AI instead of AS has been a subject of attention.
So for all db vendors, I would insist on why we suggest to use AI.

educhastenier:
Why do you have to precise it if it is already the default Bonita behaviour in PostgresInterceptor.java ?

tbouffard:
As seen with @amottier I am closing this PR as he currently has no time to work on it.
It will be reopened later.

Please do not delete the related branch to not loose the existing work.

PR #349, opened 2017-07-21
fix(security): add missing fixed bug in 7.5.0

tbouffard:
BS-15466 has been detected as fixed in 7.5.0 after the release has been
cut. So update the release notes accordingly.

Relates to [BS-15466](https://bonitasoft.atlassian.net/browse/BS-15466)


benjaminParisel (approved):
LGTM

PR #354, opened 2017-08-02
perf(engine): perf improvements in release-notes

tbouffard:
Insist on Engine startup and BDM services resume in the release notes
page to highlight this point for end-user.

Relates to [BS-11460](https://bonitasoft.atlassian.net/browse/BS-11460)

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -11,6 +11,22 @@ We have forgotten that in 7.5
 
 ### Industrialization of Living Application (in Studio)
 
+## Performance improvements
+
+### Engine startup and BPM Services resume
+
+A major improvement has been integrated in Bonita 7.6.0, especially for tenants with a lot of processes, processes with large dependencies or processes with a
+lot of dependencies
+
+Tests conducted both internally at Bonitasoft and on customer sites with real production data have shown the following decreases
+* execution time by a factor of 5 to 10
+* JVM Heap memory consumption by a factor of 2 to 3
+
+### Living Application
+
+To be documented
+
+
 ## Technology updates
 
 * The supported Tomcat version for this new release is Tomcat 8.5.16 (included in the bundle).


md/release-notes.md
```diff
@@ -11,6 +11,22 @@ We have forgotten that in 7.5
...
+A major improvement has been integrated in Bonita 7.6.0, especially for tenants with a lot of processes, processes with large dependencies or processes with a
+lot of dependencies
+
+Tests conducted both internally at Bonitasoft and on customer sites with real production data have shown the following decreases
+* execution time by a factor of 5 to 10
+* JVM Heap memory consumption by a factor of 2 to 3
+
+### Living Application
+
+To be documented
```
cpuy, line 27: I would rather document it or remove it

cpuy:
LGTM appart from the TODO part :-)

Moreover, it would be easier for us to read if you configure your IDE with 120 as line width ;-)

md/release-notes.md
tbouffard: It is a reminder to ensure we won't miss it during the release notes final review
It should be done through [BS-17011](https://bonitasoft.atlassian.net/browse/BS-17011)

tbouffard:
@cpuy thanks for your review
About the 120 line width: I did not find any information in the CONTRIBUTING.md file of the current repository. So I use a 160 line width configuration as for other Bonita projects.

We should definitely agree on the width value for doc md files and document it in the contributing guideline. 120 or 160 is ok for me.

md/release-notes.md
cpuy: Ok then I guess we can let it as it is and document it later if it is what you want

cpuy:
Actually we changed the formatter for java source code https://github.com/bonitasoft/bonita-developer-resources/pull/3

You're right there is no strict rule for md files but I believe this is a good general rule to limit to 120 for most files

tbouffard:
Ok, I didn't notice this change and I guess I am not the only one :-)

PR #360, opened 2017-08-09
feat(layout) Merge layout documentation between 7.5 and 7.6

julienmege:


md/customize-living-application-layout.md
```diff
@@ -5,18 +5,14 @@
 :::
 
 ::: warning
-Since Bonita BPM 7.3, we have made some changes in the default layout. You will have to remove `layout.css` from the page assets to make the following instructions relevant.
+Since 7.5 we made some improvements on the default layout.
+We recommand that you re-create your custom layout based on Bonita 7.5.0+ default layout.
```
abirembaut: Since this is on the 7.6 branch, I would rather say:
In 7.6 we made some improvements on the default layout.
We recommand that you re-create your custom layout based on Bonita 7.6.0+ default layout.

abirembaut (commented):
LGTM Just one comment

md/customize-living-application-layout.md
julienmege: But the 1st improvement were made in 7.5, we should: 
- Let 7.5 on every doc version 
or 
- remove version in this comment: "Since previous version  we made some improvements on the default layout."
?

abirembaut: Since the previous version we made some improvements on the default layout.
We recommand that you re-create your custom layout based on Bonita 7.6.0+ default layout.

PR #365, opened 2017-08-18
feat(doc): add in release note change in identityAPI

benjaminParisel:
Covers [BS-16519](https://bonitasoft.atlassian.net/browse/BS-16519)

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -43,7 +43,7 @@ To be documented
 ### API removals
   
 ## Breaking changes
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -43,6 +43,7 @@ To be documented
 ### API removals
   
 ## Breaking changes
+* Change in identityAPI to forbidden group name update while new name already exist.
```
danila-m: - Change in identityAPI to _forbid_ group name

- Since it will be rewritten next release anyway, might be good to add the method name here.

PR #367, opened 2017-08-22
fix(getting started tutorial): typo in table

amottier:
In version 7.3 and 7.4 we use an input widget to display a date using a date filter so in the tutorial we use to have \|date in the table content. In the update for 7.5 we use a date widget and so need to remove the filter but only the \ was removed. This PR fix the issue.


abirembaut:
LGTM

PR #368, opened 2017-08-24
fix(info): retrieving the diagram is for subscription only

danila-m:
Closes [BS-17061](https://bonitasoft.atlassian.net/browse/BS-17061)

### DIFF
--- a/md/bpm-api.md
+++ b/md/bpm-api.md
@@ -3740,7 +3740,7 @@ This method will create a new process instance ([case](#case))
       ]
     }
     ```
-### Diagram (Subscription edition only)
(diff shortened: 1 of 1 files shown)


md/bpm-api.md
```diff
@@ -3740,7 +3740,7 @@ This method will create a new process instance ([case](#case))
       ]
     }
     ```
-### Diagram
+### Diagram (Subscription edition only)
```
abirembaut: should be edition**s**

abirembaut (changes requested):
Good.
Just a s missing

abirembaut (approved):

PR #372, opened 2017-08-31
feat(process display name) Add limitation for jasper reports

abirembaut:
- provided Jasper reports still use the name rather than the
displayname. This is a known limitation.

Covers [BS-16825](https://bonitasoft.atlassian.net/browse/BS-16825)


benjaminParisel (approved):
LGTM

PR #401, opened 2017-10-23
doc(RN): document Engine deprecation of BDM APIs

educhastenier:
Closes [BS-17355](https://bonitasoft.atlassian.net/browse/BS-17355)

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -45,6 +45,8 @@ To be documented
 #### BDM Api removal
 Since 7.0, BDM APIs and model classes in packages `com.bonitasoft.engine.bdm.*` where deprecated in favor of the Community version in packages `org.bonitasoft.engine.bdm.*`.  
 They are now removed in this version.
+If you use a client Java application using Engine APIs / models in `com.bonitasoft.engine.bdm.*`, simply change the import declaration with `org.bonitasoft.engine.bdm.*`
+when upgrading to 7.6.0+.
 
 ### Configuration changes
 
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -43,9 +43,9 @@ To be documented
 ### API removals
 
 #### BDM Api removal
-Since 7.0, BDM APIs and model classes in `com.bonitasoft.engine.bdm.*` packages where deprecated in favor of the Community version in `org.bonitasoft.engine.bdm.*` packages.  
+Since 7.0, BDM APIs and model classes in `com.bonitasoft.engine.bdm.*` packages are deprecated in favor of the Community version in `org.bonitasoft.engine.bdm.*` packages.  
 They have been removed in this release.  
-If you use a client Java application using Engine APIs / models in `com.bonitasoft.engine.bdm.*`, simply change the import declaration with `org.bonitasoft.engine.bdm.*`
+If you use a client Java application depending on Engine APIs / models in `com.bonitasoft.engine.bdm.*`, simply change the import declaration with `org.bonitasoft.engine.bdm.*`
 when upgrading to 7.6.0+.
 
 ### Configuration changes
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -43,8 +43,8 @@ To be documented
 ### API removals
 
 #### BDM Api removal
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -42,6 +42,10 @@ To be documented
 
 ### API removals
 
+#### BDM Api removal
+Since 7.0, BDM APIs and model classes in packages `com.bonitasoft.engine.bdm.*` where deprecated in favor of the Community version in packages `org.bonitasoft.engine.bdm.*`.  
+They are now removed in this version.
+
```
NathalieC: What if I migrate from Bonita BPM 6.5 SP to Bonita 7.6?
Will I have anything to handle myself or can you tell me that a migration step is planned to help me handle this automatically?

educhastenier: Nothing is planned for that.
A classical user using the Studio (and not direct Engine APIs) will not have any issue, as migration to 7.0.0 requires Sam to redeploy the BDM, so the changes will be handled for them.

tbouffard: the https://documentation.bonitasoft.com/?page=migrate-from-an-earlier-version-of-bonita-bpm#toc1 migration page explicitly states that in this case, a migration step has to be managed using the Studio

tbouffard: I suggest to add a comment to point to the migration doc page as a reminder

educhastenier: I don't think this is very relevant, as someone migrating from 6.5.x to 7.6.0 will have to redeploy the BDM anyway, and will not have any problem.
More trouble would be for a customer calling Engine APIs from a client application and using deprecated model => for them, the classes will simply not be found. They will have to change the client code. Maybe we can enrich more this part...

md/release-notes.md
```diff
@@ -42,6 +42,12 @@ To be documented
 
 ### API removals
 
+#### BDM Api removal
+Since 7.0, BDM APIs and model classes in packages `com.bonitasoft.engine.bdm.*` where deprecated in favor of the Community version in packages `org.bonitasoft.engine.bdm.*`.  
+They are now removed in this version.
```
danila-m: "They have been removed in this release" might be better

md/release-notes.md
```diff
@@ -42,6 +42,12 @@ To be documented
 
 ### API removals
 
+#### BDM Api removal
+Since 7.0, BDM APIs and model classes in packages `com.bonitasoft.engine.bdm.*` where deprecated in favor of the Community version in packages `org.bonitasoft.engine.bdm.*`.
```
danila-m: "in _the `com.bonitasoft.engine.bdm.*`_ packages"

danila-m (changes requested):

md/release-notes.md
educhastenier: not sure whether both formulations are accepted...

danila-m (approved):

md/release-notes.md
```diff
@@ -42,6 +42,12 @@ To be documented
 
 ### API removals
 
+#### BDM Api removal
+Since 7.0, BDM APIs and model classes in `com.bonitasoft.engine.bdm.*` packages where deprecated in favor of the Community version in `org.bonitasoft.engine.bdm.*` packages.
```
tbouffard: "have been" instead of "where"

md/release-notes.md
```diff
@@ -42,6 +42,12 @@ To be documented
 
 ### API removals
 
+#### BDM Api removal
+Since 7.0, BDM APIs and model classes in `com.bonitasoft.engine.bdm.*` packages where deprecated in favor of the Community version in `org.bonitasoft.engine.bdm.*` packages.  
+They have been removed in this release.  
+If you use a client Java application using Engine APIs / models in `com.bonitasoft.engine.bdm.*`, simply change the import declaration with `org.bonitasoft.engine.bdm.*`
```
tbouffard: depending on instead of using

tbouffard (changes requested):
wording

md/release-notes.md
educhastenier: Nop. I say that *Since 7.0*, meaning that they were deprecated already, this is not new => simple past or present.
=> will change to **are**

md/release-notes.md
educhastenier: ok

tbouffard (approved):

PR #404, opened 2017-10-27
feat(IE11): added that we supported IE11 in 7.5

danila-m:
Closes [BS-17479](https://bonitasoft.atlassian.net/browse/BS-17479)


benjaminParisel (approved):
LGTM

PR #406, opened 2017-10-31
doc(install Studio): remove zip use with Mac

NathalieC:
remove sentence on using the zip with Mac since this has been removed from the product.


julienmege (commented):
LGTM

PR #408, opened 2017-10-31
doc(profiles): add and rework page

NathalieC:
Redo intro
Delete "create subtasks" in user
Change "custom profiles"
Place profile editor on top
Add ACME users with role "member" have all profiles


julienmege (commented):
LGTM

PR #412, opened 2017-11-03
Add info on rest api deployment

danila-m:
To better guide users to avoid bad practices.

Relates to [BS-16862](https://bonitasoft.atlassian.net/browse/BS-16862)

### DIFF
--- a/md/api-extensions.md
+++ b/md/api-extensions.md
@@ -24,8 +24,7 @@ REST API extensions use the same authorization mechanism as the standard Bonita 
 :::
 
 ::: info
-**Note**: The deployement of the REST API extensions should always be made from the portal or its APIs. Notably, the createPage() method of the PageAPI will alllow
-you to deploy a REST API extension, but will not update the security settings, meaning that you won't be able to access it at runtime. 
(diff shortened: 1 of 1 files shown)


md/api-extensions.md
```diff
@@ -23,6 +23,10 @@ REST API extensions use the same authorization mechanism as the standard Bonita
 **Note**: It is also possible to deploy a REST API extension using [REST portal API / "page" resource](portal-api.md#page).
 :::
 
+::: info
+**Note**: The deployement of the REST API extensions should always be made from the portal or its APIs. Notably, the createPage() method of the PageAPI will alllow you to deploy a REST API extension, but will not update the security settings, meaning that you won't be able to access it at runtime.
```
educhastenier: typo `deployment`

educhastenier: ...from the portal or its APIs, ***not directly from Engine APIs***. Notably, the createPage() method of the ***Engine*** PageAPI ~will~ **would** alllow you to deploy a REST API extension, but ***would*** not update the security settings, meaning that you ***wouldn't*** be able to access it at runtime.

PR #413, opened 2017-11-06
doc(SAML): add more information regarding setup with proxy

abirembaut:
- mention the name of the header (Host)
- add link to Apache reverse proxy doc
- add link to tomcat and wildfly doc


ugaston (commented):
Looks fine to me.
In step [6.] it could perhaps also be specified that the certificate to paste in the IDP section of the keycloak-saml.xml should be the certificate provided by the IdP.

abirembaut:
@ugaston : thanks for the feedback ! I just updated #414 with that.

PR #414, opened 2017-11-06
doc(SAML) add setup with certificate warning

abirembaut:
This PR adds warning mentioning that the certificate needs to be set in 2
places

Here is what the doc already mention (let me know if I can improve it) : 

- If your IdP requires the SSO requests to be signed replace the following strings in the Keys section of the SP:
  -  put your private key here
  -  put your certificate here
with you current server's private key and with the certificate provided by the IdP.

- If your IdP does not require the SSO requests to be signed, you can remove the Keys node from the SP and set the attribute signRequest to false.

- If your IdP responses are signed, replace the following strings in the Keys section of the IDP:
   - put your certificate here

- If your IdP responses are not signed, you can remove the Keys node from the IDP and set the attribute validateResponseSignature to false.

### DIFF
--- a/md/single-sign-on-with-saml.md
+++ b/md/single-sign-on-with-saml.md
@@ -152,15 +152,20 @@ For example on linux, you can use the command ssh-keygen, then go to “cd ~/.ss
     + If your IdP responses are signed, replace the following strings in the Keys section of the IDP:  
       - put your certificate here
       
+      with the certificate provided by the IdP (same certificate as in the SP section).  
       If your IdP responses are not signed, you can remove the Keys node from the IDP and set the attribute validateResponseSignature to false.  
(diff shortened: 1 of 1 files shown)


ugaston (commented):
Nice one!

Thanks.

md/single-sign-on-with-saml.md
```diff
@@ -152,15 +152,20 @@ For example on linux, you can use the command ssh-keygen, then go to “cd ~/.ss
     + If your IdP responses are signed, replace the following strings in the Keys section of the IDP:  
       - put your certificate here
       
+      with the certificate provided by the IdP (same certificate as in the SP section).
```
ugaston, line 155: Thanks, this new line definitely helps!

ugaston (approved):
Much better now. ;-)

PR #415, opened 2017-11-07
doc(SAML): fix typo loging.jsp to login.jsp

marielle-spiteri:
Fix typo in 2 sentences, (/loging.jsp) becomes (/login.jsp) : 
"Users will then be able to log in using the portal login page (/loging.jsp) provided they have a bonita account and their password is different from their username."


abirembaut:
:+1:

PR #416, opened 2017-11-09
doc(deploy) add warning on tomcat bugged version

abirembaut:
- version 8.0.32 of tomcat has an [issue](https://bz.apache.org/bugzilla/show_bug.cgi?id=58999) preventing the lib we use to compile LESS to work which make the portal unusable
Add a warning to document that

### DIFF
--- a/md/deploy-bundle.md
+++ b/md/deploy-bundle.md
@@ -42,6 +42,9 @@ Copy all files and directories from DEPLOY_ZIP_HOME/Tomcat-8.5.23/server to your
 Some configuration files from the deploy zip will overwrite some default tomcat configuration files. Proceed
 with care in a tomcat where other applications are already installed.
 :::
+:::warning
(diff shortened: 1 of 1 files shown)


md/deploy-bundle.md
```diff
@@ -42,6 +42,9 @@ Copy all files and directories from DEPLOY_ZIP_HOME/Tomcat-8.5.23/server to your
 Some configuration files from the deploy zip will overwrite some default tomcat configuration files. Proceed
 with care in a tomcat where other applications are already installed.
 :::
+:::warning
+There is an [issue on tomcat 8.0.32](https://bz.apache.org/bugzilla/show_bug.cgi?id=58999) preventing the portal Look & feel to be compiled. If you deploy bonita on an existing tomcat installation, make sure it is a different version of tomcat (preferably 8.5.x with x>=23).
+:::
```
julienmege, line 48: should we put this warning in previous versions ?

abirembaut: I don't think so as we did not support tomcat 8 before 7.6 (even though it was working)

PR #417, opened 2017-11-14
fix(profiles) hasNavigation filter in profile REST API

rbioteau:
Closes [BS-17449](https://bonitasoft.atlassian.net/browse/BS-17449)


julienmege (commented):
LGTM

PR #431, opened 2017-12-07
fix(build): update Jenkinsfile with doc site by application (removing hard-coded version)

tnguyen1:
This PR needs to be applied to all versions.


benjaminParisel (approved):
LGTM

PR #436, opened 2017-12-12
tr(merge): ignore md/release-notes.md in merge from version x to x+1

educhastenier:
Closes [BS-17700](https://bonitasoft.atlassian.net/browse/BS-17700)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+md/release-notes.md merge=ours


tbouffard (commented):
I think we should introduce this change only starting from version 7.4 as the 7.3 in the base version
Apart from that, that's sound good!

educhastenier:
Changed base version

PR #437, opened 2017-12-12
fix(doc): customization points after migration

jjargot:
With Unai we changed the content of the points dealing with bonita_home and application server customization after migration.

Versions: 7.3, 7.4, 7.5, 7.6


tnguyen1:
add to whitelist

marielle-spiteri:
I don't really like this line: "1. Configure the applicationn server at `bonita-target-version/server`: customization, libs etc."
1. Fix the typo : applicationn -> application
2. Something like that: 
"12. If you have done specific configuration and customization in your server source version, re-do it by configuring the application server at bonita-target-version/server: customization, libs etc.

educhastenier:
Looks good to me :+1:

PR #441, opened 2017-12-19
docs(release-note): add note about fix of collections modification

lcustodio:
* Add note about implication at Improvements
* Add mention at Fixes in UIDesigner component

relates to [BS-16235](https://bonitasoft.atlassian.net/browse/BS-16235)

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -83,9 +83,10 @@ When defining actors in a pool, it is now possible to remove the "initiator" tag
 ### UI Designer properties autocomplete for variables
 Autocomplete capacity has been added in UI Designer properties fields, to help you input variable names with no risk of typos anymore.
 
-### UI Designer allows handling collections with invalid state 
-Collection with invalid content now keeps the model available allowing modification on the structure of the collect (for instance add and remove). 
-This is a behaviour change that allows forms to be better managed and further custom validations possible.
+### UI Designer allows handling collections with invalid state
+Before this change a collection with invalid state has no model available. Now a collection with content in invalid state keeps the model available. 
+This allow modification on the structure of the collect with invalid state like add or remove item.
+If you make some manual validation with invalid state on collection, we advise you to check if your validations are already good.
 
 <a id="techonolgy-updates"/>
 
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -83,6 +83,10 @@ When defining actors in a pool, it is now possible to remove the "initiator" tag
 ### UI Designer properties autocomplete for variables
 Autocomplete capacity has been added in UI Designer properties fields, to help you input variable names with no risk of typos anymore.
 
+### UI Designer allows handling collections with invalid state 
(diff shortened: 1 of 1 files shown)


benjaminParisel (approved):
LGTM, @NathalieC what do you think about this note in release note ?

md/release-notes.md
```diff
@@ -83,6 +83,10 @@ When defining actors in a pool, it is now possible to remove the "initiator" tag
 ### UI Designer properties autocomplete for variables
 Autocomplete capacity has been added in UI Designer properties fields, to help you input variable names with no risk of typos anymore.
 
+### UI Designer allows handling collections with invalid state 
+Collection with invalid content now keeps the model available allowing modification on the structure of the collect (for instance add and remove).
```
NathalieC: Is it invalid content or invalid state? Article title says "state", article content says "content".
How can Fred get in this situation of invalid "state/content"? It this obvious for anyone that has not encountered the issue? If not, this should be stated.
What was the problem with the fact that the model was discarded? Fred had to create it all again or? 
The "allowing..." part of the sentence should be made into another sentence => First sentence says what has been done, second sentence talks about the value.
What is a "collect"? Collection? If so, it needs to get all its letters.

md/release-notes.md
```diff
@@ -83,6 +83,10 @@ When defining actors in a pool, it is now possible to remove the "initiator" tag
 ### UI Designer properties autocomplete for variables
 Autocomplete capacity has been added in UI Designer properties fields, to help you input variable names with no risk of typos anymore.
 
+### UI Designer allows handling collections with invalid state 
+Collection with invalid content now keeps the model available allowing modification on the structure of the collect (for instance add and remove). 
+This is a behaviour change that allows forms to be better managed and further custom validations possible.
```
NathalieC: I don't understand, sorry. 
Behaviour change? Do you mean I (as Fred) needs to do anything when I open my forms that use collections in a 7.6.1 UID or has it got no impact at all?
I don't get the value of "better management" and "further validations". Validation of what?

md/release-notes.md
```diff
@@ -205,6 +209,7 @@ has been removed.
 * BS-16996 External System Document not usable in file viewer widget or Overview
 * BS-17278 Cannot update custom widget property
 * BS-17526 Label property of input widget has not correct type
+* BS-16235 Action Remove from collection doesn't delete the correct item
```
NathalieC, line 213: The link between the bug fix and the value sold in the article is not easy to make: the item removed wasn't the right one so now we offer you to keep editing the model even if the content is invalid?
I'm not sure I'm the right person to review this PR, because lots of it seems unclear to me.
Still, if my comments make sense to you, go ahead and clarify :)

educhastenier: Do we allow ourselves to update the list of Bugs fixed in an already-released version?

lcustodio:
This change was not relevant. Release notes is already published.

PR #443, opened 2017-12-19
fix(doc): add link to section page

benjaminParisel:
Covers [BS-16698](https://bonitasoft.atlassian.net/browse/BS-16698)

Need to be re-applied into more recent version of documentation

### DIFF
--- a/md/bpm-api.md
+++ b/md/bpm-api.md
@@ -2,6 +2,8 @@
 
 ## Activities and Tasks
 
+<a id="activity"/>
+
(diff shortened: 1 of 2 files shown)


educhastenier:
why do you say `Need to be re-applied` ? merging from one version to another should do the job.

benjaminParisel:
Yes it's what i want to say

educhastenier:
so no need to say it :wink: this is the same for ALL PRs

PR #453, opened 2018-01-03
fix broken link

nchabanoles:
The link to the Engine architecture points to a non-existing page.
Replaced with a valid page name.


julienmege (commented):
👍

PR #457, opened 2018-01-18
fix(doc): wrong link to platform setup in the performance page

tbouffard:
Covers [BS-17902](https://bonitasoft.atlassian.net/browse/BS-17902)


laurentleseigneur:
verified on docker build

PR #471, opened 2018-02-15
fix error in the bdm API

alachambre:
* The API to update the bdm is  `/API/tenant/bdm`  and not `/API/accessControl/bdm`

closes [BS-18037](https://bonitasoft.atlassian.net/browse/BS-18037)


abirembaut:
👍

abirembaut (approved):

PR #472, opened 2018-02-16
fix(cluster): add missing information for AWS

tbouffard:
This change is for the 7.3 only as the hazelcast version has changed in Bonita later versions

Covers [BS-18053](https://bonitasoft.atlassian.net/browse/BS-18053)


ugaston (approved):
Look fine to me.
Thanks!

PR #473, opened 2018-02-16
fix(cluster): add missing when running inside docker

tbouffard:
As part of this fix, also remove the duplication between 'Create a
cluster from scratch' and 'Convert a single node installation into a
cluster'

Covers [BS-18053](https://bonitasoft.atlassian.net/browse/BS-18053)


ugaston (approved):
Yes!!
Nice one. Thanks again.

PR #484, opened 2018-03-01
fix(date) Add precision about time zone management into rest api

julienmege:
Closes [BS-18134](https://bonitasoft.atlassian.net/browse/BS-18134)


abirembaut (commented):
I would add something between the 2 sentences like:
"server-side we encourage you to manitpulate UTC dates only."

PR #485, opened 2018-03-01
fix(date) Add precision about time zone management into rest api (#484)

julienmege:


tbouffard (approved):
👍 as this is the same content as for the 7.6 Doc

PR #490, opened 2018-03-08
defect(API): wrong logoutService parameter description

laurentleseigneur:
* when using logoutService, redirect parameter isa query parameter, not a form param
* add information about default value


abirembaut:
:+1:

abirembaut (approved):

PR #499, opened 2018-03-22
fix(engine http client): add a warning about client/server versions

tbouffard:
Covers [BS-17624](https://bonitasoft.atlassian.net/browse/BS-17624)

### DIFF
--- a/md/configure-client-of-bonita-bpm-engine.md
+++ b/md/configure-client-of-bonita-bpm-engine.md
@@ -36,11 +36,14 @@ You can set the system property `org.bonitasoft.engine.api-type` to `HTTP`, `EJB
  * `HTTP`:
    The `HTTP` connection mode will connect the client to the engine using the http protocol. This is the most common way to connect to engine remotly.  
    When using `HTTP` mode, you must set the following system properties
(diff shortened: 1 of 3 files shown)


jjargot (commented):
I clicked on Review changes green button, is this ok?

Thank you for this documentation update.

Is it ok to add also this explanation:
"Keep in mind that after a migration, the bonita client library version used with the java application won't be aligned anymore with the new bonita engine version: plan post migration tests and possibly re compile the application."

These "post migration issues" won't occur with the HTTP REST API, no? if yes, we should recommends strongly to use REST API instead of the java API, in order to get rid of these post migration issues.

I think the migration page should also contain a new step: "if applicable, test external applications using old version of the bonita java client library, and possibly recompile using the new bonita java client library version." https://documentation.bonitasoft.com/bonita/7.6/migrate-from-an-earlier-version-of-bonita-bpm

tbouffard:
As decided with @jjargot I am going to add a note in the migration documentation page about upgrading the version of the client library in 3rd party applications

tbouffard:
@jjargot I have made a change in the migration page

jjargot (approved):
perfect for me

PR #500, opened 2018-03-26
fix(ReleaseNote): add a note about the new API method

danila-m:
addProcessComment behavior changed (some time ago in fact).
New method addProcessCommentOnBehalfOfUser() to replicate old behavior added.

Relates to : [BS-18208](https://bonitasoft.atlassian.net/browse/BS-18208)

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -21,7 +21,7 @@ modifying the Bonita installation.
 * [addProcessComment()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessComment-long-java.lang.String-) method in ProcessRuntimeAPI has had a behavior change that went unnoticed some releases ago:
 when called from a groovy script, it will systematically write the process comment as having been made by the "System" user, while previously it was using the user executing the task.
 While the new behavior will not be reverted for this method, a new method has been introduced : [addProcessCommentOnBehalfOfUser()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessCommentOnBehalfOfUser-long-java.lang.String-long-), that will allow to replicate the previous behavior of the [addProcessComment()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessComment-long-java.lang.String-) method.
-
+What it means in practice: if you are calling the method outside of groovy scripts, you can use the method you like ( addProcessComment() being probably more practical), and your process will not require any additional modification if it was migrated from an earlier version of Bonita. If you are calling the method from a groovy script, in a process designed prior to 7.3, and want to maintain the previous behavior, you will have to modify your groovy scripts to use the new API method.
 ### Jasper 5 connector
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -16,6 +16,12 @@ modifying the Bonita installation.
...
+While the new behavior will not be reverted for this method, a new method has been introduced : [addProcessCommentOnBehalfOfUser()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessCommentOnBehalfOfUser-long-java.lang.String-long-), that will allow to replicate the previous behavior of the [addProcessComment()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessComment-long-java.lang.String-) method.
+What it means in practice: if you are calling the method outside of groovy scripts, you can use the method you like ( addProcessComment() being probably more practical), and your process will not require any additional modification if it was migrated from an earlier version of Bonita. If you are calling the method from a groovy script, in a process designed prior to 7.3, and want to maintain the previous behavior, you will have to modify your groovy scripts to use the new API method.
```
NathalieC: What it means in practice if your process has been designed prior to Bonita 7.3:  
- If you are calling the method outside of groovy scripts, you can use the method you like ( addProcessComment() being probably more practical), and your process will not require any additional modification. 
- If you are calling the method from a groovy script and want to maintain the previous behavior, you will have to modify your groovy scripts to use the new API method.
- If your process has been designed in Bonita 7.4, 7.5 or 7.6, ...

md/release-notes.md
```diff
@@ -16,6 +16,12 @@ modifying the Bonita installation.
 
 ## Feature removals
 
+## API behavior change
+
+* [addProcessComment()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessComment-long-java.lang.String-) method in ProcessRuntimeAPI has had a behavior change that went unnoticed some releases ago:
```
NathalieC: "some releases ago" : "in Bonita 7.3", so we better understand why you mention 7.3 later on in the "what in means in practice" paragraph.

md/release-notes.md
```diff
@@ -16,6 +16,16 @@ modifying the Bonita installation.
 
 ## Feature removals
 
+## API behavior change
+
+* [addProcessComment()](https://documentation.bonitasoft.com/javadoc/api/7.7/org/bonitasoft/engine/api/ProcessRuntimeAPI.html#addProcessComment-long-java.lang.String-) method in ProcessRuntimeAPI has had a behavior change that went unnoticed in 7.3:
```
nchabanoles: What changed is the fix of this bug: https://bonitasoft.atlassian.net/browse/BS-14276

Operations on human tasks are asynchronous (as it should have been from 7.0.0). Hence all methods relying of Session to get the userID will find -1 as a value.

All scripts that want to perform an action on behalf of the user executing the task, should rely on the task assignee to do so. (There is no user logged hence the -1 value in the sessions; as the execution is asynchronous).

With the new API method, the script can leave a comment on behalf of the user. Note: this use case has never been considered before. As a comment was thought to be left by the user himself / herself.

nchabanoles (changes requested):

PR #501, opened 2018-03-28
fix(Portal URLs): specify autologin only supported for v6 forms

ugaston:
Versions: 7.3, 7.4, 7.5, 7.6

Specify autologin feature in 7.x is only supported for v6 instantiation forms.


abirembaut:
LGTM

ugaston:
Thanks a lot, Anthony!

PR #514, opened 2018-04-18
fix(ldap) add bonita_reactivate_users parameter

rbioteau:
Closes BS-18302


educhastenier:
Looks good to me :+1:

PR #519, opened 2018-04-25
update IAM role configuration section

JeremJR:
fix misleading section : This policy is not linked to an IAM user but to a role that we need to attach to an ec2 instance


tbouffard (approved):
LGTM so merging 👍

PR #523, opened 2018-05-03
fix(doc): Remove steps 8 to 11 only for <=7.3.1

marielle-spiteri:
Steps 8 to 11 about updating BonitaClusteredScheduler are only for version 7.3.1 or lower. 
No need to have that in page 7.4 or higher, it's confusing and user are doing the steps when they don't need. 
Migration tool takes care of it


educhastenier:
Looks good to me :+1:

PR #524, opened 2018-05-03
fix(doc): Remove steps 8 to 11 only for <=7.3.1

marielle-spiteri:
Steps 8 to 11 about updating BonitaClusteredScheduler are only for version 7.3.1 or lower. 
No need to have that in page 7.4 or higher, it's confusing and user are doing the steps when they don't need. 
Migration tool takes care of it
Removal to be effective for 7.5 and 7.6


educhastenier:
Closing this PR as there is no need to contribute a PR for all version, it will automatically merged from version 7.4 to upper btanches

marielle-spiteri:
Pages are different between 7.4 and 7.5, that is why I did two PR.  
7.5 and upper can be merged, but I believe 7.4 should be merged separately. At least that what I thought as pages are different online.

PR #529, opened 2018-05-16
doc(BDM): document usage & limitations of alternative SchemaManager

educhastenier:
* how to activate it
* warn about limitations and risks
* list error messages in case of forgetting database update manually

Closes [BS-18380](https://bonitasoft.atlassian.net/browse/BS-18380)


educhastenier:
:question: Instead of this PR, should we provide this information to Support Team by other means?

tbouffard (approved):
LGTM 👍 
This is nice to provide example of issues that can be raised in various non updated cases

PR #533, opened 2018-05-23
improvement(ldap sync): improve groups synchronization section

tbouffard:


educhastenier:
:tent: Look pretty clear to me, even though I don't understand much about LDAP...

PR #539, opened 2018-05-31
doc(BDM AC): add the make several app options

NathalieC:
* when considering the fact that several profiles can access an attribute or not, add the option of creating one app per profile, with a different page: one with the field, and one without. Play with the property hidden is not always the right choice.

### DIFF
--- a/md/bdm-access-control.md
+++ b/md/bdm-access-control.md
@@ -103,8 +103,11 @@ In the UI Designer, create an application page (*displayLeaveRequest*):
 	- **Read-only**: *true*
 
  
-Since medicalComment will not be accessible to some users, you can make its display conditional.  
-To do so, in the property **hidden** of the dedicated input, click the **fx** button to make it an expression, and write `$item.medicalComment == null || $item.medicalComment == undefined`.
+_medicalComment_ will not be accessible to the HR trainees. 
+To manage this and other differences between what profiles can access, you hove two options:
+- create one application per profile, and one page per profile: one with this field, and one without
+- manage it on the same page and make its display conditional.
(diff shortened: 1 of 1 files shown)


md/bdm-access-control.md
```diff
@@ -103,8 +103,11 @@ In the UI Designer, create an application page (*displayLeaveRequest*):
 	- **Read-only**: *true*
 
  
-Since medicalComment will not be accessible to some users, you can make its display conditional.  
-To do so, in the property **hidden** of the dedicated input, click the **fx** button to make it an expression, and write `$item.medicalComment == null || $item.medicalComment == undefined`.
+_medicalComment_ will not be accessible to the HR trainees.
```
alachambre: 2 spaces at the end, else the return line won't be done..

md/bdm-access-control.md
```diff
@@ -103,8 +103,11 @@ In the UI Designer, create an application page (*displayLeaveRequest*):
 	- **Read-only**: *true*
 
  
-Since medicalComment will not be accessible to some users, you can make its display conditional.  
-To do so, in the property **hidden** of the dedicated input, click the **fx** button to make it an expression, and write `$item.medicalComment == null || $item.medicalComment == undefined`.
+_medicalComment_ will not be accessible to the HR trainees. 
+To manage this and other differences between what profiles can access, you hove two options:
```
alachambre: type: you have

alachambre (changes requested):

NathalieC:
@alachambre ; can you please review again?

alachambre:
It's correct, but what a strange idea...
In which world is it better to have several pages with almost the same content just for a small difference when you can manage it in the same page??? It means that you will have the maintain the two pages, report all major changes twice ect ... 
Certainly not a good option in my view, but yes you can do that if you want to.

PR #563, opened 2018-07-13
fix(migration) update permission files migration procedure

abirembaut:
Migration from a version before 7.3 with custom pages and REST API is tricky because you have to report the changes to the permission files manually. It was already the case before 7.5. It's just the name of the files to edit that changed. This PR gives the right file names and procedure.
Covers [BS-18497](https://bonitasoft.atlassian.net/browse/BS-18497)


julienmege:
retest this please

PR #564, opened 2018-07-19
fix(API) Remove put payload response when does not exist

julienmege:
Related to Case Number	00021458


ugaston (approved):
Looks good to me. 
I'm guessing you tested them, so go!
Thanks,

Unai

PR #568, opened 2018-08-06
Update index before 7.8 alpha

benjaminParisel:


NathalieC:
looks good to me. Merging.

PR #570, opened 2018-08-17
fix(operations): discourage write in read-only scripts

danila-m:
Closes [BS-18701](https://bonitasoft.atlassian.net/browse/BS-18701)


tnguyen1:
retest this please

PR #577, opened 2018-09-05
Doc(kerberos): Create documentation for kerberos sso setup

abirembaut:
* create new page
* images for kerberos doc
* add reference in taxonomy.md

Covers [BS-18720](https://bonitasoft.atlassian.net/browse/BS-18720)


md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+-	Service Server: the server on which resides the service we want to access (Bonita for instance). This server is referred to as the "Principal". It uses the credentials of an AD service account and is member of the domain.
+
+-	Client: the client is the end user’s machine. The client is referred to as the "Subject". It uses the credentials of an AD user account and is member of the domain.
+
+
+## Kerberos overview for Bonita
+
+The target architecture for integrating Kerberos SSO with Bonita and Spnego will involve a custom web application filter: Bonita Kerberos filter.
+
+![Authentication over Kerberos](images/kerberos-overview.png)<!--{.img-responsive}-->
```
abirembaut, line 34: Spnego Filter should be renamed into Spnego Authenticator

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+
+The scenario covered by the filter is the following: 
+
+- The Bonita Kerberos filter intercepts all the requests to bonita portal pages, and checks if the user is already logged in on Bonita 
+    
+    - If already logged in => Allow the access
+    - If not logged in => The request is transferred to the Spnego filter
```
abirembaut: Spnego Authenticator instead of filter

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+
+The Spnego filter will then verify the user’s Kerberos tickets if present or create a new one. After obtaining a valid ticket the filter will store some information (the authenticated user login) in the client request and get back to the Bonita Kerberos filter.
```
abirembaut: Spnego authenticator

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+	
+- The Bonita Kerberos filter will automatically create a Bonita session and let the user through to access the Portal resources.
+	
+:::info 
+**Note:** The Spnego filter is called programmatically by the Bonita Kerberos filter. It is not deployed as a standard web application filter in the application descriptor (web.xml). This is done in order to allow bi-directional exchanges with the Bonita Kerberos filter (regular web filters are only chained in one direction).
```
abirembaut: This note can be removed as it was true for the old implementation but it no longer applies

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+1- Obtain from your AD administrator the AD service account that is going to be used as the SPN. If none exist yet, follow the steps below else, ignore this.
+
+2- Check that the service account is not already associated to any host by running this command (replace bonita.tomcat with your service account name):
+
+```
+setspn -L bonita.tomcat
+```
+This will list the hosts associated to this service account. There should not be any entry.
+
+3- Run these two commands in order to declare the SPN (bonita.tomcat). The first command will use the simple host name (win2008tomcat):
```
abirembaut, line 73: is it always win2008tomcat ?

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+   edit the authenticationManager-config.properties as follows:
+   
+```
+		# saml.logout.global = false
+		# auth.tenant.admin.username = install
+		# auth.passphrase = BonitaBPM
+   
+	-->	auth.AuthenticationManager = org.bonitasoft.console.common.server.auth.impl.kerberos.RemoteAuthenticationManagerImpl
+	-->	kerberos.filter.active = true
+	-->	kerberos.auth.standard.allowed = true
```
abirembaut: maybe we should set it to false
I will change it in the engine too

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+		# auth.AuthenticationManager = org.bonitasoft.console.common.server.auth.impl.oauth.OAuthAuthenticationManagerImpl
+		# OAuth.serviceProvider = LinkedIn
+		# OAuth.consumerKey = ove2vcdjptar
+   -->  logout.link.hidden=true 
+```
+
+Make sure to [set the right tenant admin username](multi-tenancy-and-tenant-configuration#toc2).
+It is recommended to also replace the value of the passphrase (property auth.passphrase) which is used by the engine to verify the authentication request.
+The value must be the same as in the file **bonita-tenant-sp-custom.properties**.  
+If some users need to bypass kerberos authentication method, you can authorize it by setting the property `kerberos.auth.standard.allowed` to true. Users will then be able to log in using the portal login page (/login.jsp) provided they have a bonita account and their password is different from their username.
```
abirembaut: I think we should remove the "some" (If some users need to bypass) and add a sentence:
If only a limited group of users need to bypass kerberos authentication method you can restrain it by setting the property `kerberos.auth.standard.allowed` to false and setting the property `auth.tenant.standard.whitelist` with the list of authorized usernames (coma separated).

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+
+5. In the tenant_portal folder of each existing tenant: `<BUNDLE_HOME>/setup/platform_conf/current/tenants/<TENANT_ID>/tenant_portal`,
+   edit the spnego-config.properties file as follows:
+   
+```
+	-->  spnego.allow.basic          = true
+	-->	 spnego.allow.localhost      = true
+	-->	 spnego.allow.unsecure.basic = true
+	-->	 spnego.login.client.module  = spnego-client
+	-->	 spnego.krb5.conf            = krb5.conf
```
abirembaut: conf/krb5.conf

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+	-->	 spnego.login.conf           = login.conf
```
abirembaut: conf/login.conf

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+		#authenticator.delegate=casAuthenticatorDelegate
+		#authentication.delegate.cas.server.url.prefix=http://ip_address:port
+		#authentication.delegate.cas.service.url=http://ip_address:port/bonita/loginservice
+```
+  
+It is recommended to also replace the value of the passphrase (property auth.passphrase). The value must be the same as in the file **authenticationManager-config.properties** updated previously.
+
+8. If your Domain Controller is correctly configured, you are done.  
+Then you can try to access a portal page, an app page or a form URL (or just `http://<host>:<port>/bonita[?tenant=<tenantId>]`) and make sure that you are automatically logged in.  
+Note that if you try to access `http://<bundle host>:<port>/bonita/login.jsp`, then you won't be redirected as this page still needs to be accessible in order for the tenant administrator (or another user if you set the property `saml.auth.standard.allowed` to true) to be able to log in without an account on AD.
```
abirembaut: or `auth.tenant.standard.whitelist`

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+
+## LDAP synchronizer and Kerberos
+
+If you are using an LDAP (or AD) service and the [LDAP synchronizer](ldap-synchronizer.md) to manage your user data,   
+you can continue to do this and manage authentication over Kerberos.  
+The LDAP synchronizer user must be registered in Bonita (no need for an LDAP/AD account). It is recommended though to use the tenant admin account.   
+We recommend that you use LDAP or AD as your master source for information, synchronizing the relevant information with your Bonita platform.
+
+::: info
+**Note :** By default the [LDAP synchronizer](ldap-synchronizer.md) sets the password of the accounts created with the same value as the username. So, even if you allow standard authentication (by setting the property `saml.auth.standard.allowed` in **authenticationManager-config.properties**), users won't be able to log in with the portal login page directly without going through the IdP.
```
abirembaut: y without going through the Domain Controller instead of IdP ?

md/single-sign-on-with-kerberos.md
```diff
@@ -0,0 +1,331 @@
...
+The LDAP synchronizer user must be registered in Bonita (no need for an LDAP/AD account). It is recommended though to use the tenant admin account.   
+We recommend that you use LDAP or AD as your master source for information, synchronizing the relevant information with your Bonita platform.
+
+::: info
+**Note :** By default the [LDAP synchronizer](ldap-synchronizer.md) sets the password of the accounts created with the same value as the username. So, even if you allow standard authentication (by setting the property `saml.auth.standard.allowed` in **authenticationManager-config.properties**), users won't be able to log in with the portal login page directly without going through the IdP.   
+:::
+
+## Single sign-on with Kerberos using the REST API
+
+Kerberos is a browser-oriented protocol (based on http automatic redirection, forms, etc...), therefore only resources that require a direct access from a web browser are handled by the Kerberos filter.
```
abirembaut: We should remove everything that's before "only resources that"
Because I don't think Kerberos is a browser-oriented protocol. Only our Kerberos filter is browser oriented

abirembaut (commented):
Just a few comments.
Other than than it looks good to me

abirembaut:
Also missing: Wildfly specifics (Work In Progress)

md/single-sign-on-with-kerberos.md
mehdi-kettani: no, it depends on what you use as hostname of your service machine.

md/single-sign-on-with-kerberos.md
mehdi-kettani: i'll do the change on the image later.

abirembaut: :+1:

md/single-sign-on-with-kerberos.md
abirembaut: :+1:

benjaminParisel (approved):

PR #580, opened 2018-09-12
feat(howtos): bonita as a windows service

ecuriotto:
Tutorial to show how to install Bonita with Tomcat as a windows service
versions: 7.4,7.5,7.6,7.7,7.8


tnguyen1:
add to whitelist

md/taxonomy.md
```diff
@@ -257,5 +258,4 @@
   * [Map authorization rules](custom-authorization-rule-mapping.md)
   * [Use the rich textarea widget in a mail template](rta-mail-template.md)
   * [Manage dates in BDM and User Interfaces](dates-management-tutorial.md)
-
-
+  * [Install Tomcat with Bonita as a service in Windows](bonita-as-windows-service.md)
```
tbouffard: I have just discussed with @NathalieC about the taxonomy
We think a better place for this is in the `Platform installation examples` part of the `Installation` section.
How-tos are for Bonita usage
❓ what do you think about this change?

md/bonita-as-windows-service.md
```diff
@@ -0,0 +1,100 @@
...
+* Postgresql Database (or another database supported by Bonita)
+Please verify the version supported from the Support page
+
+## Download and installation
+
+* **_Download_** apache-tomcat-x.y.exe from Tomcat website: https://tomcat.apache.org/
+* **_Install_** apache-tomcat-x.y.exe. From now on we'll call the installation path as %TOMCAT_INSTALL_FOLDER%
+* **_Verify_** that Tomcat is well installed browsing the following URL: http://localhost:8080
+
+![Tomcat home page](images/bonita-as-windows-service/tomcatHome.png)
```
tbouffard, line 21: ⚠️ the image is very large, can it be reduced?
in addition, the version shown here in not the one supported for Bonita 7.4

md/bonita-as-windows-service.md
```diff
@@ -0,0 +1,100 @@
...
+If you are a Windows server administrator you know how useful can be to have softwares installed as a windows service.
+In this tutorial we show how to install Tomcat with Bonita configured as a windows service.
+
+## Software needed
+
+For this tutorial you will need the following softwares:
+* Tomcat
+* Bonita deploy Bundle
+* Postgresql Database (or another database supported by Bonita)
+Please verify the version supported from the Support page
```
tbouffard: ⚠️ missing break line
⚠️ wording:  should be `supported version` instead of `version supported `

tbouffard (changes requested):
As a tutorial, this page is nice (I made some remarks which are minor)
However, I suggest to move the page in the taxonomy and to rethink about the page purpose: there are a lot of information which are duplicated with the bundle installation page and which are not presented in the same order which may be confusing and duplication add maintenance.
I understand that for a tutorial this is easier to have all information on the same page, but here, I think this would be easier to say: follow the bundle installation and put emphasys on what is specific to the Tomcat service configuration

md/taxonomy.md
ecuriotto: Sorry for the late answer, I think it's ok in the Platform installation examples. I'm just lacking time to complete...

md/bonita-as-windows-service.md
ecuriotto: how to do to have different images for the different product versions? Maybe we can just update the image during resizing for removing the version, leave only:
image

ecuriotto: the image has to be resized, I agree :-)

ecuriotto: I have removed the version in the image

md/bonita-as-windows-service.md
ecuriotto: I have changed "version supported" in "supported version" and added a break line

tbouffard (approved):

tbouffard:
Thanks @ecuriotto for integration the review feedbacks.
So I am merging this PR

PR #581, opened 2018-09-12
fix(Languages): add warning for translation of Mobile Portal

ugaston:
Versions: 7.4, 7.5, 7.6, 7.7

Add a warning on Languages: when adding a new language to the Portal, if the new language is to be used in the Mobile Portal, a few precautions must be taken for the Mobile portal to display correctly.

Relates to BS-18494


NathalieC:
relates to [BS-18494](https://bonitasoft.atlassian.net/browse/BS-18494)

PR #584, opened 2018-09-17
emphase the warning on bad API usage

philippelaumay:
add warning and mitigation for document and business objects


philippelaumay:
could you apply these modifications from 7.3

PR #585, opened 2018-09-17
update and promote new components

philippelaumay:
updates to push BCD capabilities instead of workspaceAPI. 
also BDM capabilities instead of custom data type


philippelaumay:
could you apply these modifications from 7.3 version

PR #592, opened 2018-09-27
doc(cluster):Make cluster installation consistent

nchabanoles:
When converting a single node installation into a cluster we must make sure to disable the 2nd level cache to avoid data consistency errors


educhastenier (approved):
Looks good to me :+1:

educhastenier:
I'll rebase it onto 7.4 before merge.

PR #594, opened 2018-10-02
 fix(log): Fix typo User profile with U upper case

marielle-spiteri:
searchOptionsBuilder.filter(ProfileSearchDescriptor.NAME, "User");


md/manage-users.md
```diff
@@ -25,10 +25,10 @@ final User user2 = identityAPI.createUser(creator);
 Now add the user to a Bonita BPM Portal profile. A user who does not have a profile cannot log in to Bonita BPM Portal.
 ```groovy
 // reference the user in the profile User
-// the user must be now registered in one profile. Let's choose the profile user
+// the user must be now registered in one profile. Let's choose the profile User
```
educhastenier: to be more English-style, I would switch word order to:
`Let's choose the "User" profile`

PR #597, opened 2018-10-11
fix(Studio Preferences): document new Tomcat JVM args settings

ugaston:
Version: 7.8

Document the new Studio Preferences that allow to set the Tomcat's JVM args.
Also, do some cleaning, as many of the documented settings are no longer there in the Studio Preferences. (Presumably, they have not been there for a few versions, now... So this cleaning should be backported to the documentation of previous versions.)

Relates to BS-18942


educhastenier:
Looks good to me :+1:

PR #599, opened 2018-10-12
fix(Case API) document case API supported filters

abirembaut:
Covers [BS-17993](https://bonitasoft.atlassian.net/browse/BS-17993)


benjaminParisel (approved):
LGTM

PR #607, opened 2018-10-25
feat(renaming) use project terminology

rbioteau:
Update terminology.
`repository` -> `project` when talking about a bonita repository.
A repository stay a repository when talking about Git, maven or svn repository.


md/share-a-project-on-github.md
```diff
@@ -1,11 +1,11 @@
-# How to share a repository on Github 
+# How to share a project on Github
```
educhastenier: I would not rename the file, allowing to switch from one version of Bonita to another, staying on the same 'page'

alachambre: is it a frequent use case for customers ?

educhastenier: When on a wrong version, you click in the menu corresponding to your version, if the page has the same name, you stay on the page, which is convenient.
So yes.

rbioteau: What about search and seo ? Does it indexes the article name ?

NathalieC: Switching version and staying on the same page is a use case.
Keeping repository becomes an issue.
=> Switch to "project" n the URL, and create a redirect when coming from the same page on previous versions? How complex?

educhastenier: I suggest to keep the same file name.
When changing from BonitaBPM to Bonita, we agreed to keep the same file names, because of SEO and because of bookmarking. Why should we do different this time?

NathalieC: It is the backup solution that we will adopt if the redirect is too much of a pain.
I'll let you set the complexity.

NathalieC: final decision is not to change the md name.

PR #608, opened 2018-10-26
feat(event): remove ARCHIVED_FLOWNODE_INSTANCE event

baptistemesta:
This event will not be thrown in 7.8 by default.
To be merged along with bonitasoft/bonita-engine-sp/pull/1096


educhastenier:
:question: Did you check the doc is up to date with the code?

baptistemesta:
I did not checked all elements, but they should not have changed

PR #622, opened 2018-11-13
fix(widgets): remove widget after container

NathalieC:
* remove "widget" in the titles "form container widget" and "tab container widget".
* add "a special kind of widget" to the description of container


abirembaut:
LGTM

PR #623, opened 2018-11-16
fix(SAML SSO): correct and improve configuration instructions

ugaston:
Versions: 7.5, 7.6, 7.7, 7.8

Try to make the configuration of the keycloak-saml.xml a bit clearer, especially for the cases where signature/encryption is required.


abirembaut:
Awesome !
Good job !
Just a comment: I don't think we need to display the full stacktrace in the troubleshooting section as it takes a lot of space. The first 2 lines of the trace and the "caused by" line are clear enough to troubleshoot

I will make another PR to update the [keycloak-saml.xml](https://github.com/bonitasoft/bonita-engine-sp/blob/master/subscription/platform/platform-resources-sp/src/main/resources/tenant_template_portal/keycloak-saml.xml) accordingly

abirembaut (approved):
LGTM

ugaston:
Cheers, Anthony!

PR #624, opened 2018-11-16
feat(rel-notes): add performance improvements

NathalieC:
*add performance improvements

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -114,12 +114,20 @@ In this Bonita 7.8 release, two pages have been rewritten:
 <a id="performance"/>
 
 ### Performance
-A few improvements have been brought to the overall Bonita efficiency:
-- Deleting an archived case went down from seconds to milliseconds. Particularly enjoyable on environments with a large amount of archived data. Moreover, deleting cases in batches is now nearly as fast as deleting a single case and is not constrained by database connexion timeouts. 
+The following improvements have been made to Bonita performance:
 - There is less I/O requests to temp folders
-- Works execution uses less database resources
+- Work execution consumes less database resources
 - The login REST APIs mechanism consumes less resources
-- At engine startup and BPM services resume, the process dependencies loading requires less memory
+- At engine startup and BPM services resume, the process dependencies loading consumes less memory
+
+A special attention has also been carried out to an improved deletion mechanism of archived cases. Deleting archived cases is now way more efficient.
+
+:::warning
+As a result, some [Events](event-handlers.md) are not triggered anymore (among others: ARCHIVED\_FLOWNODE\_INSTANCE\_DELETED, only concerns deletion of **archived** elements).
+If you used these events and still need them, there is a way to switch back to the previous deletion mechanism.
+In such cases, please contact Customer Support to activate this legacy deletion mechanism.
+Be aware that this legacy deletion mechanism is deprecated and will be deleted in a future version. There will be not support in the long-term.
+:::
 
 <a id="rest-timeout"/>
 
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -98,7 +98,7 @@ A new theme is available for any Living Application you create: "bonita-default-
 
 <a id="uuid"/>
 
-### Stability of files throughout revisions
+### Stability of sources files throughout revisions
 Updating timers, business data, data dependencies, messages, pools, contract inputs, parameters, expressions, connectors, and documents, used to create additional UUID (Universal Unique Identifier) that resulted in false positives when comparing two revisions of the same file (using Git or SVN). This has been improved to make only real changes stick out when using a Diff Tool.
 
 <a id="gwt"/>
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
  - **BPM Services** page for the profile Technical Administrator
    It has been created using the UI Designer and the new modal container. Since it is dedicated to a profile critical for the good initialization and management of Bonita, the page has not been made available in the portal's resources. It cannot be added to any application nor be edited in the UI Designer. It's designed has been slightly reviewed compared to the GWT one.
    
+<a id="performance"/>
+
+### Performance
+A few improvements have been made to the engine efficiency:
+- Deleting an archived case now takes only a few milliseconds
+- The number of accesses to temp folders has been reduced on the file system to limit I/O access 
+- Works execution uses less database resources
+- The login REST APIs mechanism consumes less resources
+- At engine startup and BPM services resume, jars are written on disk right after retrieving them from the database to save memory 
+
 <a id="rest-timeout"/>
 
 ### REST connector Read timeout is configurable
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
  - **BPM Services** page for the profile Technical Administrator
    It has been created using the UI Designer and the new modal container. Since it is dedicated to a profile critical for the good initialization and management of Bonita, the page has not been made available in the portal's resources. It cannot be added to any application nor be edited in the UI Designer. It's designed has been slightly reviewed compared to the GWT one.
    
+<a id="performance"/>
+
+### Performance
+A few improvements have been made to the engine efficiency:
+- Deleting an archived case now takes only a few milliseconds
```
tbouffard: ℹ️ seems maybe a too big commitment (few milliseconds) but we can say that deletion is done in a factor of milliseconds whereas in previous Bonita versions, this was done with a factor of seconds
We could also insist on the fact that
* this is particularly visible on environments with a large amount of archived data
* batch deletion
  * is almost as fast as single deletion. Previously the time increased in proportion of the number of the cases to be deleted
  * was then constrained by the overall timeout process (mainly database connection timeout) so its usage was limited only to environments with medium dataset
  * now provides more maintenance facilities because it is faster and better works on large environments

md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
  - **BPM Services** page for the profile Technical Administrator
    It has been created using the UI Designer and the new modal container. Since it is dedicated to a profile critical for the good initialization and management of Bonita, the page has not been made available in the portal's resources. It cannot be added to any application nor be edited in the UI Designer. It's designed has been slightly reviewed compared to the GWT one.
    
+<a id="performance"/>
+
+### Performance
+A few improvements have been made to the engine efficiency:
```
tbouffard: ℹ️ I would not put emphasis on the engine part as even most of the improvements have been done on that component, Portal parts have also been improved (profiles and login)

tbouffard:
```suggestion
Improvements have been made in the following area:
```

md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
...
+- Deleting an archived case now takes only a few milliseconds
+- The number of accesses to temp folders has been reduced on the file system to limit I/O access
```
tbouffard:
```suggestion
- There is less I/O requests to temp folders
```

tbouffard: I am very confident with the wording with may ask someone else

md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
...
    
+<a id="performance"/>
+
+### Performance
+A few improvements have been made to the engine efficiency:
+- Deleting an archived case now takes only a few milliseconds
+- The number of accesses to temp folders has been reduced on the file system to limit I/O access 
+- Works execution uses less database resources
+- The login REST APIs mechanism consumes less resources
+- At engine startup and BPM services resume, jars are written on disk right after retrieving them from the database to save memory
```
tbouffard: ℹ️ I would add the impacted use case in addition to technical explanation. I am not sure this is releavant to dig into so much technical details. I think that the result is more important
Let's talk about it directly

tbouffard:
```suggestion
- At engine startup and BPM services resume, the process dependencies loading requires less memory
```

tbouffard (changes requested):
Let's talk directly about my comments 😺

md/release-notes.md
NathalieC: made a summary

md/release-notes.md
NathalieC: removed. thx

md/release-notes.md
NathalieC: I'm taking your words in the meantime

md/release-notes.md
NathalieC: thx

md/release-notes.md
NathalieC: changed into "overall Bonita" efficiency

md/release-notes.md
NathalieC: There are many customers encountering the issues you have fixed. 
Non-technical people will not dig in but just see there are many areas of improvements.
Technical people will notice the potential amount of improvement, that we ourselves cannot easily quantify.

md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
  - **BPM Services** page for the profile Technical Administrator
    It has been created using the UI Designer and the new modal container. Since it is dedicated to a profile critical for the good initialization and management of Bonita, the page has not been made available in the portal's resources. It cannot be added to any application nor be edited in the UI Designer. It's designed has been slightly reviewed compared to the GWT one.
    
+<a id="performance"/>
+
+### Performance
+A few improvements have been brought to the overall Bonita efficiency:
```
tbouffard: ℹ️ can we have the same wording in 7.7.5 and 7.8.0?

md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
...
+- Deleting an archived case went down from seconds to milliseconds. Particularly enjoyable on environments with a large amount of archived data. Moreover, deleting cases in batches is now nearly as fast as deleting a single case and is not constrained by database connexion timeouts.
```
tbouffard: ℹ️ we should change this as the improvement is only partial for now (see what we document for 7.7.5)

md/release-notes.md
```diff
@@ -111,6 +111,16 @@ In this Bonita 7.8 release, two pages have been rewritten:
...
+<a id="performance"/>
+
+### Performance
+A few improvements have been brought to the overall Bonita efficiency:
+- Deleting an archived case went down from seconds to milliseconds. Particularly enjoyable on environments with a large amount of archived data. Moreover, deleting cases in batches is now nearly as fast as deleting a single case and is not constrained by database connexion timeouts. 
+- There is less I/O requests to temp folders
+- Works execution uses less database resources
+- The login REST APIs mechanism consumes less resources
+- At engine startup and BPM services resume, the process dependencies loading requires less memory
+
```
tbouffard, line 131: ⚠️ special warning about should be added about the triggered events change
See what we have in 7.7.5

tbouffard (changes requested):

md/release-notes.md
NathalieC: ok. Changed.

md/release-notes.md
NathalieC: ok. Changed.

md/release-notes.md
NathalieC: OK. Copied and pasted.

md/release-notes.md
tbouffard: Looks ok now 😄

tbouffard (approved):

PR #625, opened 2018-11-19
feat(7.7.5): add RN for 7.7.5 release

educhastenier:
+ describe with more details the performance improvements

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -84,7 +84,24 @@ Bonita binaries as well as Studio installer are now officially signed on Windows
 ### Form mapping API error messages
 Improvements of messages in case of error on form mapping API calls have been done to explain the cause of error and give options for easier error recoveries.
 
-<a id="techonolgy-updates"/>
+### Performance (applies to Bonita 7.7.5+)
+The following improvements have been made to Bonita performance:
+- There is less I/O requests to temp folders
+- Work execution consumes less database resources
+- The login REST APIs mechanism consumes less resources
+- At engine startup and BPM services resume, the process dependencies loading consumes less memory
+
+A special attention has also been carried out to an improved deletion mechanism of archived cases. Deleting archived cases is now way more efficient.
+
+:::warning
+As a result, some [Events](event-handlers.md) are not triggered anymore (among others: ARCHIVED\_FLOWNODE\_INSTANCE\_DELETED, only concerns deletion of **archived** elements).
+If you used these events and still need them, there is a way to switch back to the previous deletion mechanism.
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -84,7 +84,23 @@ Bonita binaries as well as Studio installer are now officially signed on Windows
...
 
-<a id="techonolgy-updates"/>
+### Performance (applies to Bonita 7.7.5+)
+The following improvements have been made to Bonita performance:
+- There is less I/O requests to temp folders
+- Work execution consumes less database resources
+- The login REST APIs mechanism consumes less resources
+- At engine startup and BPM services resume, the process dependencies loading consumes less memory
+
+A special attention has also been carried out to an improved deletion mechanism of archived cases. Deleting archived cases is now way more efficient.
```
NathalieC, line 94: We should make it homogeneous with the chapter on 7.8.0.
Current proposition (waiting for review):
A few improvements have been brought to the overall Bonita efficiency:
- Deleting an archived case went down from seconds to milliseconds. Particularly enjoyable on environments with a large amount of archived data. Moreover, deleting cases in batches is now nearly as fast as deleting a single case and is not constrained by database connexion timeouts. 
- There is less I/O requests to temp folders
- Works execution uses less database resources
- The login REST APIs mechanism consumes less resources
- At engine startup and BPM services resume, the process dependencies loading requires less memory

md/release-notes.md
```diff
@@ -84,7 +84,23 @@ Bonita binaries as well as Studio installer are now officially signed on Windows
...
+
+:::warning
```
NathalieC, line 96: I thought we were about deprecating and deleting those event handlers. 
So there will be no long term option of reactivating it back.
The value (or lack of value) of these events should be pointed here so the reader realizes there is no point of wanting them.
I have not synchronized with CS before writing that. If you have, then your words must be kept.

NathalieC: When we have an agreement on that, this must be added to the 7.8.0 release notes too.

NathalieC:
More reviewers are needed to check my point of view.

md/release-notes.md
educhastenier: We plan to have the same content in 7.8, instead of the existing PR

md/release-notes.md
educhastenier: It will be reported to 7.8.0 RN, as soon as CS confirmes your remark above.

educhastenier: Checked and integrated review from CS team...

PR #635, opened 2018-12-04
fix(SAML) nameID and IDP entityID configuration

abirembaut:
- Add nameIDPolicyFormat (mandatory for AD FS)
- explain that the IDP entityID should be updated

Covers [BS-19158](https://bonitasoft.atlassian.net/browse/BS-19158)


ecuriotto:
Does the nameIDPolicyFormat need to be insert also when the PrincipalNameMapping policy="FROM_ATTRIBUTE" is set?
If the answer is yes then everything is fine, otherwise we may maybe expain when to insert this attribute or not

abirembaut:
yes, it should be present all the time

PR #637, opened 2018-12-05
fix(SAML) be more specific for unsigned responses

abirembaut:


md/single-sign-on-with-saml.md
```diff
@@ -190,7 +190,7 @@ _If your IdP does neither require the SSO requests to be signed nor encrypt its
 :::
 
 :::info
-_If your IdP responses are not signed, you can remove the Keys node from the IDP and set the attributes validateRequestSignature and validateResponseSignature to false._
+_If your IdP responses are not signed, you can remove the Keys node and the attributes signaturesRequired and signatureAlgorithm from the IDP and set the attributes validateRequestSignature and validateResponseSignature to false._
```
ugaston, line 193: Well, I don't know about that....

According to the Keycloak [doc](https://www.keycloak.org/docs/latest/securing_apps/index.html#_saml-general-config):
> signaturesRequired: If set to true, the client adapter will sign every document it sends to the IDP. Also, the client will expect that the IDP will be signing any documents sent to it. (...)

So supposing you could have SP requests signed and IdP responses unsigned, you would still want to keep both _signaturesRequired_ and _signatureAlgorithm_....

abirembaut: Ah, yes. You are right !

PR #646, opened 2019-01-15
fix(bar) process image not generated in bar

rbioteau:
Closes BS-19258


md/release-notes.md
```diff
@@ -148,6 +148,8 @@ In practice, it means that if your process has been designed prior to Bonita 7.3
 ### Deprecated Workspace API
 The Workspace API tooling (headless studio build) has been deprecated. You are recommended to use the *LA builder* which is part of the tooling suite of [Bonita Continuous Delivery add-on](https://documentation.bonitasoft.com/bcd/2.0/).
 
+### Business Archive content updated
```
amottier: @rbioteau maybe add: "(*.bar)" after Business Archive to clarify what we are referring to.
Also I would write: Business ARchive or Business archive but not Business Archive (yes I'm picky).

PR #647, opened 2019-01-15
fix(matrix) clarify java supported version

laurentleseigneur:
could be understood as "... java 9,10,11 are supported"
remove "and above" note about java 8

need to be merged in later doc versions


tbouffard (approved):
Nice!

PR #648, opened 2019-01-16
fix(connector): typo in method list

laurentleseigneur:
* remove extra `>`


educhastenier:
Is it only on 7.8?

laurentleseigneur:
you're right, it's from 7.3 a least, will change base branch

PR #656, opened 2019-01-24
fix(migtool-chglog): update rel. date for 2.37.0

NathalieC:
* change 17th to 23rd as the release date


amottier:
@NathalieC as the support of a new version of Bonita is actually a change added to a new version of the migration tool I would add it as an item of the list just as a normal item (i.e. same as a bug fix or another improvement).

As the list of changes is always short I think it would not be useful to separate bug fixes from improvements.

Anyway if you prefer current version I'm also fine to merge this as is.

NathalieC:
I only had the BS numbers on the pdf where I took all the information from past versions of the migration tool.
I don't want to spend time to look into Jira for each one to find their issue type.
So we'll leave bug fixes and improvements merged for now.
I'll split them in the future versions.

educhastenier:
Looks good to me :+1:

PR #657, opened 2019-01-29
rn(java11): groovy lib upgrade

tbouffard:
Covers [BS-19278](https://bonitasoft.atlassian.net/browse/BS-19278)


educhastenier:
Looks good to me :+1:

PR #666, opened 2019-02-08
feat(release): added logger for work execution audit

baptistemesta:
Relates to [BS-18482](https://bonitasoft.atlassian.net/browse/BS-18482) and [BR-21](https://bonitasoft.atlassian.net/browse/BR-21)

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -85,3 +85,12 @@ The version of the _javax.mail:mail_ dependency has been updated from _1.4.5_ to
 #### Twitter connector
 
 The version of the _org.twitter4j:twitter4j-core_ dependency has been updated from _4.0.2_ to _4.0.7_
+
+## Monitoring capabilities
+
+### Engine work execution
+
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -33,3 +33,10 @@ In order to be compatible with Java 11, you must update the following plug-ins d
 
 The following Bonita dependencies have been upgraded to improve the Java 11 support
 * groovy-all from 2.4.4 to 2.4.16
+
+## Monitoring capabilities
+
+# Engine work execution
+
+A new logger has been added to detect long running work execution.
```
tbouffard: ℹ️ I would put a link to the new page created as part of #667 and remove the part about activation.
We should also explain the value added like we do in the configuration page

For instance (to be reviewed with the PO)
```
In Bonita 7.9, a work execution audit mechanism has been introduced. It can be activated
to detect when a work takes too much time to be executed or it was _rescheduled_ too much
times.
```

tbouffard (approved):

PR #668, opened 2019-02-14
fix(Define ... BDM): replace default relationship

antiablanco:
* replace default relationship from composition to aggregation

Please repeat same operation for v7.8


NathalieC:
I'll repeat this in 7.8. Thanks.

PR #700, opened 2019-03-22
Give more details about the default transition on inclusive gateway

alachambre:
[BST-151](https://bonitasoft.atlassian.net/browse/BST-151)

### DIFF
--- a/md/gateways.md
+++ b/md/gateways.md
@@ -22,4 +22,8 @@ several outputs defined, a condition is required to determine which one fires.
   ![diagram of a inclusive gateway](images/images-6_0/papde_pm_diag_gateways_inclusive_gate.png)
 
 An inclusive gateway waits for the input from all active paths, and activates all outgoing transitions that have a condition that
-evaluates to true. If there is only one outgoing transition, no condition is needed and the transition is always activated. If there is more than one outgoing transition, every transition must either be the default or must have a condition.
+evaluates to true.  
+If there is only one outgoing transition, no condition is needed and the transition is always activated.  
+If there is more than one outgoing transition, every transition must either be the default or must have a condition.
+The default transition is used if all other conditions are evaluated to false. It is not mandatory to have a default transition, but it is a way to ensure that your process won’t end successfully in the middle of its flow because all conditions returned false.  
+If you don't want to define a default transition, you should ensure that at least one of your conditions will be evaludated to true.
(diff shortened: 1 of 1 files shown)


md/gateways.md
```diff
@@ -22,4 +22,8 @@ several outputs defined, a condition is required to determine which one fires.
   ![diagram of a inclusive gateway](images/images-6_0/papde_pm_diag_gateways_inclusive_gate.png)
 
 An inclusive gateway waits for the input from all active paths, and activates all outgoing transitions that have a condition that
-evaluates to true. If there is only one outgoing transition, no condition is needed and the transition is always activated. If there is more than one outgoing transition, every transition must either be the default or must have a condition.
+evaluates to true.  
+If there is only one outgoing transition, no condition is needed and the transition is always activated.  
+If there is more than one outgoing transition, every transition must either be the default or must have a condition.  
+The default transition is used if all other conditions are evaluated to false. It is not mandatory to have a default transition, but it is a way to ensure that your process won’t end successfully in the middle of its flow because all conditions returned false.  
+If you don't want to define a default transition, you should ensure that at least one of your conditions will be evaludated to true.
```
baptistemesta:
```suggestion
If you don't want to define a default transition, you should ensure that at least one of your conditions will be evaluated to true.
```

md/gateways.md
```diff
@@ -22,4 +22,8 @@ several outputs defined, a condition is required to determine which one fires.
   ![diagram of a inclusive gateway](images/images-6_0/papde_pm_diag_gateways_inclusive_gate.png)
 
 An inclusive gateway waits for the input from all active paths, and activates all outgoing transitions that have a condition that
-evaluates to true. If there is only one outgoing transition, no condition is needed and the transition is always activated. If there is more than one outgoing transition, every transition must either be the default or must have a condition.
+evaluates to true.  
+If there is only one outgoing transition, no condition is needed and the transition is always activated.  
+If there is more than one outgoing transition, every transition must either be the default or must have a condition.
```
baptistemesta:
```suggestion
If there is more than one outgoing transition, every transition should either be the default or should have a condition.  
```

md/gateways.md
baptistemesta: After reading again the code:

parallel gateway: 
* all non default transitions are taken (without checking conditions), default transition is not taken (???)

inclusive:
* all transition that have a condition evaluated to "true" or DO NOT HAVE CONDITION (considered as tue) are taken
* if there is none the default transition is taken
* if there is no default an **exception is thrown**

exclusive:
* first transition that have a condition evaluated to "true" or DO NOT HAVE CONDITION (considered as tue) is taken
* if there is none the default transition is taken
* if there is no default an **exception is thrown**

alachambre: Parallel -> You can't add condition or set as default an outgoing transition of a parallel gateway from the Studio -> all transitions are taken -> make sens

I'll add some details according to your informations, thx

PR #717, opened 2019-04-16
ROI(LDAP): Move link to server side doc

danila-m:
Seen with Harris, the way it is now needlessly confuses new users.
Moved the link accordingly.


harissubasic (approved):
OK to remove the link

PR #727, opened 2019-05-02
fix(deployment): bonita portal as remote client of the bonita engine

tbouffard:
The previous way of document this was mixed with how to configure custom
code being client of the Bonita Engine.
In addition, provide explanation for both the Tomcat and Wildfly bundles

### DIFF
--- a/md/configure-client-of-bonita-bpm-engine.md
+++ b/md/configure-client-of-bonita-bpm-engine.md
@@ -4,6 +4,10 @@ If you want your application to use the engine API of a remote engine, your appl
 
 The engine (the *server*) must also be configured to accept the connection.
 
+**Note**: the Bonita Portal can be configured to be used as a Client of the Bonita Engine. See the [Alternate Bonita Deployment](two-main-types-of-deployment.md)
+documentation for more information
+
+
 <a id="client_config" />
 
 ## Client configuration
@@ -12,12 +16,6 @@ There is three ways to configure the client to connect to the Bonita Engine.
 
 First the engine client verify if the configuration is set programmatically,
 then it tries to get the configuration set in system properties, finally it falls back to the legacy way of configuration that is the bonita home client.
-
-::: info
-In order to be able to start the portal with a remote engine the following system properties must be set in the JVM starting the Portal:
- * `org.bonitasoft.platform.username` must be set with the username of the platform admin
- * `org.bonitasoft.platform.password` must be set with the password of the platform admin
-:::
 
 ### Configure Client using programmatic configuration
 You can configure the connection to the engine directly using `org.bonitasoft.engine.util.APITypeManager`, please refer to the [Javadoc](http://documentation.bonitasoft.com/javadoc/api/${varVersion}/org/bonitasoft/engine/util/APITypeManager.html).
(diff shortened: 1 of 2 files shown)


tbouffard:
**Important**: modifications must be applied to the 7.9 backport in the Tomcat bundle Bonita Portal configuration as we have switched from Bitronix to Arjuna/Narayana
The introduction must also be changed as there is no more `deploy` bundle

md/configure-client-of-bonita-bpm-engine.md
```diff
@@ -4,6 +4,10 @@ If you want your application to use the engine API of a remote engine, your appl
 
 The engine (the *server*) must also be configured to accept the connection.
 
+**Note**: the Bonita Portal can be configured to be used as a Client of the Bonita Engine. See the [Alternate Bonita Deployment](two-main-types-of-deployment.md)
```
ecuriotto, line 7: Is it just the Bonita Portal or all the web layer?

md/two-main-types-of-deployment.md
```diff
@@ -50,7 +45,150 @@ But you can still use the pre packaged Tomcat bundles or the `bonita.war` webapp
...
+```
+
+On Windows (start-bonita.bat)
+```
+   call setup\setup.bat init %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
+   if errorlevel 1 (
+       goto exit
+   )
+```
+
```
ecuriotto: Should we add in this paragraph to remove the setup folder?

md/two-main-types-of-deployment.md
```diff
@@ -50,7 +45,150 @@ But you can still use the pre packaged Tomcat bundles or the `bonita.war` webapp
...
+        </datasources>
+    </subsystem>
+```
+
+
+Edit the `<bonita-installation-directory>/server/standalone/deployments/bonita-all-in-one-x.y.z.ear/bonita.war/WEB-INF/web.xml`
+
+Apply the same changes as for the Tomcat bundle to disable Engine resources
+
+
```
ecuriotto, line 183: In the "Configuring the Bonita Stack" chapter there's no mention to changes in the server with the engine. Is it correct?

md/two-main-types-of-deployment.md
```diff
@@ -26,20 +26,15 @@ It is fast because the Bonita Portal and the Bonita Engine run on the same JVM a
...
-
-::: info
-Don't forget to set in this JVM the following properties:
- * `org.bonitasoft.platform.username` must be set with the username of the platform admin
- * `org.bonitasoft.platform.password` must be set with the password of the platform admin
-:::
-
 ### Accessible through HTTP
 
 ![deploy2](images/images-6_0/poss_deploy2.png)
```
ecuriotto, line 38: It should be good to add to the picture a reference to the two application servers and then use it in the chapter. If we use "Bonita portal", this term should appear in the picture

ecuriotto (commented):
Very good, following these instructions I could setup a working environment with client and engine in separate servers

md/configure-client-of-bonita-bpm-engine.md
tbouffard: In the documentation, we only talk about the Portal.

md/two-main-types-of-deployment.md
tbouffard: I don't understand `portal` is already present in all pictures of this page

image

md/two-main-types-of-deployment.md
tbouffard: Yes, that's correct
I will add a section about this to make this explicit

md/two-main-types-of-deployment.md
tbouffard: I document to use the setup tool to configure the Bonita Portal bundle but for now this is only for the setenv file
I have to check how we document the alternative authentication configuration which is a mix between server side and Bonita configuration (which requires configuration in the setup tool only used with the Engine part here)

tbouffard: done with an additional commit

tbouffard:
Merging to propose the solution that has been validated by @ecuriotto 
Feel free to open new pull request to improve this documentation

PR #739, opened 2019-05-16
feat(documents) Tutorial to create a document validation process

alachambre:
It is a good exemple to illustrate how managing documents with Bonita 7.9 is Handy.

[BST-261](https://bonitasoft.atlassian.net/browse/BST-261)


rbioteau:
retest this please

md/list-of-documents.md
```diff
@@ -1,9 +1,42 @@
...
+:::
+
+This exemple shows how to easily create a document validation process using Bonita Studio.  
+A document validation process is a process where a User should be able to first submit a list fo documents. Then, an other user can perform a review on the list submitted, and ask for some changes on the list. If the reviewer asks for it, then the submitter has to edit the list of documents by adding / editing / removing some documents of the list, according to the review.  
+
+Here is a preview of the process we are going to build:  
+![Validation process](images/documentValidationProcess.png)
+
+ 1. Create a process _documentValidation_, with two lanes: _Submitter lane_ and _Reviewer lane_. 
+ 2. Create also two actors: _Submitter actor_ and _Reviewer actor_. Each actor should be mapped to a different part of the organization (two different users for exemple). Set the _Submitter actor_ as initiator, and map each lane to its actor.
```
nchabanoles: :information_source: I would have kept Submitter and Reviewer. I do not find any value in naming an actor "actor".

md/list-of-documents.md
```diff
@@ -1,9 +1,42 @@
...
+ 3. Create a multiple document variable named _documents_, and a process variable of type boolean named _submissionValid_.
+ 4. Select the pool and create a contract for the data _documents_.  ⚠️ Make sure to use the button **Add from data** to create your contract, else the contract input is not bounded to the data, you will have to code many things by yourself.
```
nchabanoles: I think it is better to explain the value of doing the action instead of stating that the other option is difficult.
:x: "else the contract input is not bounded to the data, you will have to code many things by yourself" ==> "the tool will automatically generate the operations to update the data when the task will be submitted by the user"

It is an example of wording, feel free to adapt it to make it easier to understand by the user.

md/list-of-documents.md
```diff
@@ -1,9 +1,42 @@
...
+ 5. Create the instantiation form using the Studio shortcut. Rename the form into _initDocuments_, and then go back to the Studio.
+ 6. Select the task _Review_. The objective of this task is to display the documents submitted to the reviewer. He should then be able to validate or not the submission, depending on the documents. To do so, we are going to use a small trick to gain some development time.
```
nchabanoles: :information_source: the trick shows us that there is a need for data management in the UID. Let's do it in the next release and come back on this tutorial to amend it.

md/list-of-documents.md
```diff
@@ -1,9 +1,42 @@
...
+    - Create a contract input for the document data _documents_, using the button Add from data. This contract input will be removed in the second time, but it will allow us to generate a form which display the documents (that's the trick).
+    - Create an other contract input (using the button add this time) of type BOOLEAN, named _review_.
+    - Create the form using the Studio shortcut. Rename this form review, **and remove the file upload widget from the form** -> The form display the documents, and give the possibility to validate the submission. Save and go back to the Studio.
+    - On the task review, remove the contract input for the documents (the form has been generated, we do not need it anymore). 
+    - On the operations panel, remove the operation that update the list of documents (same idea), and add an operation to assign the value of the contract input _review_ to the process data _submitionValid_. At the end, you should have only one operation: _submitionValid_ takes value of _review_.
```
nchabanoles: :question: is this trick really saving time? I think the text forgtet to mention that the formOutput should also be updated to stop sending the documents...

:information_source: I think I would personnaly prefer that the tutorial explains me how to create the form without the trick. (this would be more helpful in other situations when I will need repeated containers)

md/list-of-documents.md
nchabanoles: :warning: It is rarely a good idea to map an actor to named users, it is always better to map to a role, group or membership. That is why I would rather not suggest to do it, in a tutorial.

nchabanoles (changes requested):
Reviewed made directly from GH, I did not deploy it to validate the rendering.

md/list-of-documents.md
alachambre: easier to find all actors when you use the Studio search (my opinion)

md/list-of-documents.md
alachambre: ok
It saves time if you know what you are doing, but if so you do not really need a tutorial ^^'

md/list-of-documents.md
alachambre: ok

PR #765, opened 2019-06-07
feat(release notes): add bug fixes

NathalieC:
* add bug fixes for UI D
more to come

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -1,9 +1,5 @@
 # Release notes
 
-::: info
-**Note:** The 7.9 is currently work in progress (WIP). The 7.9.0 GA is planned on June 2019.
-:::
-
 ## New values added
 
 <a id="improve-form-generation"/>
@@ -366,8 +362,6 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 
 ## Bug fixes
 
-### Fixes in Documentation
-
 ### Fixes in Bonita 7.9.0 (2019-06-13)
 #### Fixes in Engine component
 * BR-69 SQLServerException error when deleting cases with more then 2100 subprocesses
@@ -387,6 +381,7 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 * BS-19322 Contract validation fail if a complex contract input is null and has a multiple complex child contract input
 * BS-19335 Timers triggers keep being rescheduled and are not fired on time or fired randomly
 * BS-19345 Script Connector does not return stderr in connector result
+
 #### Fixes in Studio component
 * BST-126 Studio must be restarted to use User-defined functions
 * BST-132 Wrong script generated in some specific cases with multiple data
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -364,4 +364,21 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
-(more to come on June)
+## Bug fixes
+
+### Fixes in Documentation
+
+### Fixes in Bonita 7.9.0 (2019-06-13)
+#### Fixes in Engine component
+
+#### Fixes in Studio component
+
```
pabloalonso, line 393:
```suggestion
* BST-126 Studio must be restarted to use User-defined functions
* BST-132 Wrong script generated in some specific cases with multiple data
* BST-162 Generate process documentation does not work post 7.7
* BST-204 Default Application created in the Portal raises an error about custompage_home when imported and Deployed from Studio
* BST-221 UID allows to have 2 forms with same name when created from the Studio: this breaks LA Builder
* BST-271 Import .bpmn diagram fails
```

md/release-notes.md
```diff
@@ -364,4 +364,21 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
 ## Bug fixes
 
-(more to come on June)
+## Bug fixes
+
+### Fixes in Documentation
+
+### Fixes in Bonita 7.9.0 (2019-06-13)
+#### Fixes in Engine component
+
```
pabloalonso, line 385:
```suggestion
* BR-69 SQLServerException error when deleting cases with more then 2100 subprocesses
* BS-15896 Unable to configure EntityManagerFactory
* BS-18364 Connector Replay stays in state TO_RE_EXECUTE when executeFlowNode arrives at the same time the task is transitioning from 'initializing' to 'ready'
* BS-18734 Deploying a BAR with two forms with the same name but a different case leads to Unique index or primary key violation CONSTRAINT_INDEX_29 BAR_RESOURCE(TENANTID, PROCESS_ID, NAME, TYPE)
* BS-18748 Portal Organization IMPORT button has no feedback
* BS-18874 work execution: 5 seconds SELECT * FROM flownode_instance WHERE kind = 'boundaryEvent' ...
* BS-18880 work execution: 84 seconds SELECT * FROM from flownode_instance where logicalGroup3=nn and terminal=0 order by id ASC limit 1;
* BS-18910 message_instance x waiting_event couples: 72 seconds request duration
* BS-19082 Long execution duration for this processAPI.searchAssignedAndPendingHumanTasks API call
* BS-19239 Exception during Quartz Job execution leaves the associated flownode in WAITING state and the process execution is stopped
* BS-19293 REST API extension: ClassCastException
* BS-19295 SELECT widget sets the field of the variable holding the initial value to null when the page/form loads if it is shared by another SELECT widget
* BS-19298 Process may end up locked forever
* BS-19319 [BDM] Business Object Java Setter does not create aggregated instances
* BS-19322 Contract validation fail if a complex contract input is null and has a multiple complex child contract input
* BS-19335 Timers triggers keep being rescheduled and are not fired on time or fired randomly
* BS-19345 Script Connector does not return stderr in connector result
```

md/release-notes.md
```diff
@@ -364,4 +364,21 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
+#### Fixes in UI Designer component
+* UID-29 RichTextArea readOnly property should be dynamic
+* UID-34 ui-bootstrap-tpls.js asset should be displayed in the Assets panel for the modal container (as for tabs container)
+* UID-69 Localization help section is empty in custom widget editor
+* UID-86 Modal container: the focus should be located on the input element
+* UID-90 Deleting one asset from a custom Widget based on the default one remove all the assets
+
+#### Fixes in Web/Portal component
+
```
pabloalonso:
```suggestion
* BPO-71 REST API - bpm/humanTask filter by displayName doesn't work
```

pabloalonso (changes requested):

PR #768, opened 2019-06-11
Groovy in Bonita

rbioteau:


tnguyen1:
retest this please

tnguyen1:
retest this please

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+```groovy
+userInput as UserApp
+```
+
+### List Business Objects using DAO
+
+When defining your BDM you can write custom queries in JPQL than can be called using the object DAO.
+Some queries are generated by default like: `find`, `findByFirstName`, `findByLastName`...
+
+All object DAO are injected at runtime in Groovy script expression or can be retrieved using [APIClient#getDAO](https://documentation.bonitasoft.com/javadoc/api/7.9/org/bonitasoft/engine/api/APIClient.html#getDAO-java.lang.Class-)
```
educhastenier: Please use `${varVersion}` instead of `7.9`

rbioteau: ❤️

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
+# Groovy script language in Bonita
+
+In Bonita, it is a prerequisite to be comfortable with Java when implementing advanced behaviors in your processes, the below article aims at helping with Groovy, the scripting language used by Bonita.
+Groovy has been chosen as the scripting language because it is :
+* Concise, readable and have an expressive syntax, easy to learn for Java developers
+* Seamlessly and transparently integrates and interoperates with Java and any third-party libraries
```
alachambre, line 6: Almost true, you can't use native java lambda syntax, you must transform it to a closure.

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+userInput COMPLEX
+	firstName TEXT
+	lastName TEXT
+```
+
+As a COMPLEX input is stored as Map it is possible to write
+
+```groovy
+userInput as UserApp
+```
```
alachambre, line 247: I don't really understand this part:
* In your model, the class is AppUser and not UserApp
* I thought this was an exemple of _how to retrieve a complex child from its parent_, so I was expecting something like `commentInput.appUser as AppUser` ?

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+
+### List Business Objects using DAO
+
+When defining your BDM you can write custom queries in JPQL than can be called using the object DAO.
```
alachambre:
```suggestion
When defining your BDM you can write custom queries in JPQL that can be called using the object DAO.
```

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+Some queries are generated by default like: `find`, `findByFirstName`, `findByLastName`...
+
+All object DAO are injected at runtime in Groovy script expression or can be retrieved using [APIClient#getDAO](https://documentation.bonitasoft.com/javadoc/api/${varVersion}/org/bonitasoft/engine/api/APIClient.html#getDAO-java.lang.Class-)
+
+Here is the UserAppDAO interface generated for the UserApp object:
+
+```java
+public interface UserAppDAO extends BusinessObjectDAO {
```
alachambre: Again, you defined an object AppUser in your model

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Update a Business Data
+
+The example below is the generated code when editing a UserApp address from a contract
+
+`user` is the existing data in the process  
+`userInput` is the contract input of the edition task
+
+```groovy
+if (!userInput?.address) { // As Address is not mandatory it can be null
```
alachambre, line 293: Maybe a beginner needs to know that in Groovy, Object boolean evaluation is different that in java: 
_A null object reference evaluates to false, a non null reference to true, but for for strings, collections, iterators and enumerations it also takes into account the size._
-> It is the first boolean evaluation in this tutorial, but maybe a line about this important Groovy behavior in the general section would be nice?

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+	return null
+}
+def addressVar = user.address ?: new com.company.model.Address() // Retrieve the existing address or create a new one
+addressVar.street = { //Use a Closure to resolve the street value
+	if (!userInput?.address?.street) { // Street is mandatory so it can't be null here, protected by a contract constraint, null-check statement is generated anyway
+		return null
+	}
+	def streetVar = addressVar.street ?: new com.company.model.Street() // Retrieve the existing street or create a new one
+	// Assign contract values, note that nullsafe navigators are used even if we know that userInput.address is not null here
```
alachambre, line 302:
```suggestion
	// Assign contract values, note that nullsafe operators are used even if we know that userInput.address is not null here
```

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+As it is generated code it has to work in many situation so it is not the most concise code.
+Here is another example where it updates the _lastEditedBy_ aggregation relation  
+
+```groovy
+//Retrieve aggregated UserApp using its DAO and persistenceId
+def userAppVar = userAppDAO.findByPersistenceId(commentInput?.lastEditedBy?.persistenceId_string?.trim() ? commentInput.lastEditedBy.persistenceId_string.toLong() : null)
+if (!userAppVar) { // no userApp found for the given persistenceId
+	if (commentInput?.lastEditedBy?.persistenceId_string?.trim() ? commentInput.lastEditedBy.persistenceId_string.toLong() : null) {
```
alachambre, line 319: Maybe a word about the `trim()` operation ?

alachambre (changes requested):
Looks good!
* There is some inconsistencies between the model you define and your exemples that should be fixed
* I also added some suggestions here and there ..

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,347 @@
...
+* Concise, readable and have an expressive syntax, easy to learn for Java developers
+* Seamlessly and transparently integrates and interoperates with Java and any third-party libraries
+
+
+## From Java to Groovy
+
+For a complete description, please have a look at the reference documentation [here](http://groovy-lang.org/differences.html).
+
+### Strings
+
```
educhastenier, line 14: Is it our role to explain what groovy is and what are the language capabilities?

alachambre: We gave our point of view on that, product owner decide.

md/groovy-in-bonita.md
rbioteau: the Complex input used is describe above

md/groovy-in-bonita.md
rbioteau: yes, I'll add section about Groovy Thruth

md/groovy-in-bonita.md
```diff
@@ -0,0 +1,359 @@
...
+def ternaryOutput = (sampleText != null) ? sampleText : 'Hello Groovy!'
+ 
+// The Elvis operator in action. We must read: 'If sampleText is not null assign
+// sampleText to elvisOuput, otherwise assign 'Viva Las Vegas!' to elvisOutput.
+def elvisOutput = sampleText ?: 'Viva Las Vegas!'
+```
+
+### Groovy truth
+
+Groovy decides whether an expression is true or false by applying the rules given below.
```
alachambre, line 172: the following rules doesn't apply when using the null safe operator, it only check if the reference is null or not.

rbioteau: I don't understand what you mean ?

alachambre:
``` groovy
def emptyString = ""
println emptyString?.toLong()
// NumberFormatException -> the check on the String doesn't return false, toLong() is performed.

if (emptyString) {
    println emptyString.toLong()
} else {
    println "nothing to print"
}
// `nohting to print` -> if(emptyString) return false, groovy truth apply

```

rbioteau: It is a border use case.

PR #779, opened 2019-07-01
Fix getting-started-tutorial.md, radio button values/displayed keys

bflorat:
fix doc on radio button values/displayed keys.
Without quotes on 'displayed keys' and 'returned keys', I get : 
```
[]  {"label":"Approve request","value":"approved"}
[] {"label":"Refuse request","value":"refused"} 
```

### DIFF
--- a/md/getting-started-tutorial.md
+++ b/md/getting-started-tutorial.md
@@ -480,8 +480,8 @@ Now define the form for the _Manager review_ task. Start by automatically genera
     | --------- | ------ |
     | Label  | Review decision  |
     | Available values (click on the "fx" binding icon next to the input field first) | statusValues  |
-    | Displayed keys  | label  |
-    | Returned keys  | value  |
+    | Displayed keys  | "label"  |
+    | Returned keys  | "value"  |
     | Selected Value  | `formInput.status`  |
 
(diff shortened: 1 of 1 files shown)


tbouffard:
@amottier can you please review this PR and also check if this should be applied to older documentation version

amottier:
Actually I review the instruction and did some tests and I think the current indications in the documentation are correct but maybe misleading. It is expected that radio button values/displayed keys are configured using constant and not dynamic value (i.e. not need to click on the fx button). If constant is used no need for quotes. Dynamic value is only needed for the available values.
If dynamic value is selected for values/displayed keys then having quote becomes mandatory to define  them as strings instead of variables names.
I create an example to illustrate (that can be imported and preview in the UI Designer) the two valid configuration and the one that failed: https://www.dropbox.com/s/vqcn02oyz4bxzrp/form-testRadiobButton.zip?dl=0
@bflorat if you agree I'll not merge this PR. I'm in a process of rewriting the getting started tutorial and I'll make sure to clarify this point and include a screenshot of expected widget settings.

bflorat:
Thanks for the explanations, I'll close the PR.

PR #784, opened 2019-07-05
fix(doc): fix incorrect configuration keys

educhastenier:
fix the key prefix for all archive configuration keys

Closes [BS-19370](https://bonitasoft.atlassian.net/browse/BS-19370)


educhastenier:
:warning: seems weird we never saw it before !

PR #795, opened 2019-07-30
add(release notes): 7.9.2 first iteration

NathalieC:
* add 7.9.2 bug fixes
As fix version is missing on some bug fixes, the list may evolve with the information I get during the day.

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -434,7 +434,7 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 * BST-126 Studio must be restarted to use User-defined functions
 * BST-132 Wrong script generated in some specific cases with multiple data
 * BST-162 Generate process documentation does not work post 7.7
-* BST-201 Contract input name should never be equals to the associated business data name
+* BST-201 Contract input name should never be equal to the associated business data name
 * BST-202 Application descriptor Deploy button does not trigger the saved changes and doesn't raise errors
 * BST-204 Default Application created in the Portal raises an error about custompage_home when imported and Deployed from Studio
 * BST-221 UI Designer allows to have 2 forms with same name when created from the Studio: this breaks LA Builder
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -364,6 +364,23 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 ## Bug fixes
 ### Fixes in Documentation
 * BS-19323 Bonita documentation is not clear enough about PostgreSQL versions supported
+* BST-151 Default outgoing transition not executed when  there is more than one outgoing transition from inclusive gateway
+* BST-269 Document workaround for macos - ssh issue
+* BST-171 Can't share a project on github using HTTPS and double authentication factor
+* BST-341 On Mac, Studio does not start if recent Java versions installed
+
+### Fixes in Bonita 7.9.2 (2019-08-01)
+#### Fixes in Engine component
+* BS-19299 GET REST API connector raised an exception if answer HTTP code is 204 or 404
+
+#### Fixes in Studio component
+* BST-345 Open an invalid access control file leads to an NPE instead of displaying the `error view`
+* BST-349 The option to initialize a business data using a Query is not always available
+* BST-365 BS-19299 - GET REST API connector raised an exception if answer HTTP code is 204 or 404
+
+#### Fixes in Web/Portal component
+* BPO-24 BS-18487 cannot download a document from the FilePreview widget link
+* BPO-189 Case overview page for an archived case is not working in IE 11 (id not found)
 
 ### Fixes in Bonita 7.9.1 (2019-07-04)
 #### Fixes in Studio component
@@ -376,6 +393,7 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 * BST-299 BPMN export fails if Data to send in Call Activity contains empty association
 
 #### Fixes in UI Designer component
+* UID-82 When a new Fragment is created, the Palette is not updated
 * UID-83 Message toast in UI Designer is not clear when license is expired
 * UID-104 Fragment variable value isn't clean when variable scope change
 * UID-105 When name of artifact is too long, left panel is broken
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -364,6 +364,22 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 ## Bug fixes
 ### Fixes in Documentation
 * BS-19323 Bonita documentation is not clear enough about PostgreSQL versions supported
+* BST-151 Default outgoing transition not executed when  there is more than one outgoing transition from inclusive gateway
+* BST-269 Document workaround for macos - ssh issue
+* BST-171 Can't share a project on github using HTTPS and double authentication factor
+* BST-341 On Mac, Studio does not start if recent Java versions installed
```
alachambre: Would you mind to change the description of the BST-341 ? As far as we know the encountered issue is more related to JRE configurations than to `recent Java versions installed`, I think it would be a terrible idea to write this in the release note ...

alachambre:
```suggestion
* BST-341 On Mac, Studio installer might not detect the installed JRE
```

NathalieC: Sure.

md/release-notes.md
```diff
@@ -364,6 +364,22 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
+
+### Fixes in Bonita 7.9.2 (2019-08-01)
+#### Fixes in Engine component
+* BS-19299 GET REST API connector raised an exception if answer HTTP code is 204 or 404
```
alachambre: The fix has been done in the connector not in the Engine

NathalieC: We usually do not have such a component to report bug fixing. Either Studio or Engine.

alachambre: And the exact description would be `GET REST API connector raised an exception if answer body is null` (which is often the case when the code is 204)

md/release-notes.md
```diff
@@ -364,6 +364,22 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
+
+#### Fixes in Studio component
+* BST-345 Open an invalid access control file leads to an NPE instead of displaying the `error view`
+* BST-349 The option to initialize a business data using a Query is not always available
+* BST-365 BS-19299 - GET REST API connector raised an exception if answer HTTP code is 204 or 404
```
alachambre: Duplication, there were two tickets because I was afraid to forget to integrate the new connector version in the Studio. 
I suggest you to write something like 'Bonita Studio now uses REST API connector 1.0.6`

NathalieC: thanks for noticing
I've deleted the duplicate and added a technical update chapter in a higher release notes content

md/release-notes.md
educhastenier: @alachambre The procedure is more "you change the Issue title and then you notify that it has changed", so that the release notes and the Jira title are consistent

md/release-notes.md
```diff
@@ -364,6 +364,22 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 ## Bug fixes
 ### Fixes in Documentation
 * BS-19323 Bonita documentation is not clear enough about PostgreSQL versions supported
+* BST-151 Default outgoing transition not executed when  there is more than one outgoing transition from inclusive gateway
```
educhastenier:
```suggestion
* BST-151 Default outgoing transition not executed when there is more than one outgoing transition from inclusive gateway
```

md/release-notes.md
```diff
@@ -364,6 +364,22 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
+* BST-269 Document workaround for macos - ssh issue
+* BST-171 Can't share a project on github using HTTPS and double authentication factor
+* BST-341 On Mac, Studio installer might not detect the installed JRE
```
educhastenier:
```suggestion
* BST-341 On MacOS, Studio installer might not detect the installed JRE
```

md/release-notes.md
```diff
@@ -364,6 +364,22 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 ## Bug fixes
 ### Fixes in Documentation
 * BS-19323 Bonita documentation is not clear enough about PostgreSQL versions supported
+* BST-151 Default outgoing transition not executed when  there is more than one outgoing transition from inclusive gateway
+* BST-269 Document workaround for macos - ssh issue
```
educhastenier:
```suggestion
* BST-269 Document workaround for MacOS - ssh issue
```

md/release-notes.md
```diff
@@ -415,16 +432,24 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
 * BST-126 Studio must be restarted to use User-defined functions
 * BST-132 Wrong script generated in some specific cases with multiple data
 * BST-162 Generate process documentation does not work post 7.7
+* BST-201 Contract input name should never be equals to the associated business data name
```
educhastenier:
```suggestion
* BST-201 Contract input name should never be equal to the associated business data name
```

md/release-notes.md
```diff
@@ -415,16 +432,24 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
+* BST-202 Application descriptor Deploy button does not trigger the saved changes and doesn't raise errors
```
educhastenier:
```suggestion
* BST-202 Application descriptor Deploy button does not trigger the saved changes and does not raise errors
```

md/release-notes.md
```diff
@@ -415,16 +432,24 @@ Please use the Tomcat bundle instead, or see the [Custom Deployment](deploy-bund
...
 * BST-204 Default Application created in the Portal raises an error about custompage_home when imported and Deployed from Studio
 * BST-221 UI Designer allows to have 2 forms with same name when created from the Studio: this breaks LA Builder
+* BST-251 manage organizations wizard display issue
```
educhastenier: not precise enough. What was the issue?

md/release-notes.md
educhastenier: Does it still make sense to separate Portal, Engine, Connectors, ...?
Can't we just say Runtime env (We + Portal + Engine) / Development env (Studio + UID) ?
That sounds more meaningful to my understanding of what a Bonita Customer might expect...

md/release-notes.md
alachambre: A table  sometime very small...

alachambre: But it has been fixed like 4 months ago, what does this have to do in the 7.9.2 release note?

NathalieC: If you look carefully, it's in the 7.9.0 section: at the time of 7.9.0, many bug fixes had no fixversion.
They have been added recently, so I'm catching up on the release notes from 7.9.0.
Not mentionning the 7.8.x bug fixes that have a pretty vague fixversion (like 7.8.x....).
We need to do better on fix versions to avoid such issues.

md/release-notes.md
NathalieC: This question will be addressed for 7.10. I'm taking notes.

md/release-notes.md
NathalieC: changed in Jira and release notes to "On MacOS, manage organizations wizard needs scrolling to add a membership"

pabloalonso (approved):

PR #810, opened 2019-08-30
fix(doc): fix incorrect information about BDM

Ilagouilly:
Correct the default value of the BDM relationship between new objects (parent and child) from composition to aggregation.
Versions: 7.7, 7.8, 7.9


tnguyen1:
retest this please

danila-m (approved):

PR #815, opened 2019-09-06
fix(setup remote engine): use template files

laurentleseigneur:
* use setup folder templates files instead of tomcat/wildfly ones

* add support for ...
* fix ... behaviour
* refactor ....

### DIFF
--- a/md/two-main-types-of-deployment.md
+++ b/md/two-main-types-of-deployment.md
@@ -87,7 +87,7 @@ Remove the content of the `setup` directory as the setup tool in not used on the
 
 ##### Tomcat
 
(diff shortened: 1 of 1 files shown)


tbouffard (changes requested):
👎 the 1st step of the procedure is to remove the setup tool
So that's why we request to update files in the `server` folder

Note that the choice of removing the setup tool folder has been done after reviews with @ecuriotto : this highly simplifies the understanding for the end user. The setup tool is not required so we do not used any of its resource for the installation.
I have just discussed today with @pabloalonso about this topic. Feel free to contact him for more information and evenutal improvements in that area in the next Bonita versions

laurentleseigneur:
ok with setup tool removal.

PR #816, opened 2019-09-06
feat(ssl): use TLS v1.2

laurentleseigneur:
TLS v1 as shown in example is deprecated (see https://www.chromestatus.com/feature/5654791610957824)
TLS v1.3 is not yet supported without activating specific options in browsers

TLS v1.2 has been tested on a regular Bonita tomcat bundle with chrome and firefox


laurentleseigneur:
replaced by https://github.com/bonitasoft/bonita-doc/pull/824

PR #817, opened 2019-09-09
fix(getting-started): fix `info` typos and broken UI Designer icon

tnguyen1:
fix(getting-started): fix `info` typos and broken UI Designer icon


amottier:
Thanks for fixing the typo!

PR #826, opened 2019-09-13
fix(requirements): specific supported OS versions

tbouffard:
Describe the only tested and supported versions. Bonita could eventually work with newer versions
but we cannot commit ourselves on it.
This removal is also done to avoid misunderstanding with end users.

For instance, Bonita 7.6 has been released on December 2017 and Windows Server 2019 on November 2018: at the Bonita release time, we didn't know that Bonita would work with this unreleased Windows version.


ugaston (commented):
This looks fine to me.

Do you think the same should be applied to Red Hat and Ubuntu?

tbouffard:
@ugaston yes you are right
I am going to also update the page accordingly

ugaston (approved):
Fine by me.
Thanks.

PR #829, opened 2019-09-18
fat(http): add info on http client connections

baptistemesta:
The java client in http mode has now a configurable maximum number of
connections. Add that to the documentation

Relates to [BS-19395](https://bonitasoft.atlassian.net/browse/BS-19395)

### DIFF
--- a/md/configure-client-of-bonita-bpm-engine.md
+++ b/md/configure-client-of-bonita-bpm-engine.md
@@ -37,7 +37,7 @@ You can set the system property `org.bonitasoft.engine.api-type` to `HTTP`, `EJB
    * `org.bonitasoft.engine.api-type.server.url`: it is the url of the server, e.g. for a engine on the same host it can be `http://localhost:8080`
    * `org.bonitasoft.engine.api-type.application.name`: it is the name of the web aplication on wich the engine HTTP API is deployed, e.g. `bonita`
 
-  In addition the number of connections used by the client can be configured with system property `org.bonitasoft.engine.api-type.connections.max`, the default is 20.
+  In addition, starting from version 7.9.4, the number of connections used by the client can be configured with system property `org.bonitasoft.engine.api-type.connections.max`, the default is 20.
 
 ::: warning
 We do not guarantee to keep the http protocol stable, so we strongly recommend that you use the same version for both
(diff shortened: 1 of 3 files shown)


julienmege (commented):
👍

educhastenier:
:warning: if no 7.9.4 is ever released, change base of this PR to 7.10 !

md/configure-client-of-bonita-bpm-engine.md
```diff
@@ -37,6 +37,8 @@ You can set the system property `org.bonitasoft.engine.api-type` to `HTTP`, `EJB
    * `org.bonitasoft.engine.api-type.server.url`: it is the url of the server, e.g. for a engine on the same host it can be `http://localhost:8080`
    * `org.bonitasoft.engine.api-type.application.name`: it is the name of the web aplication on wich the engine HTTP API is deployed, e.g. `bonita`
 
+  In addition the number of connections used by the client can be configured with system property `org.bonitasoft.engine.api-type.connections.max`, the default is 20.
```
tbouffard: ⚠️ I suggest to state this is only available as of Bonita 7.9.4
Otherwise people using 7.9.3 may think that this option is available

md/configure-client-of-bonita-bpm-engine.md
```diff
@@ -73,6 +75,7 @@ org.bonitasoft.engine.api-type = LOCAL
 #org.bonitasoft.engine.api-type = HTTP
 #server.url = http://localhost:8080
 #application.name = bonita
+#connections.max = 20
```
educhastenier, line 78: :question: did you check it works with the Bonita Home client file configuration?

md/configure-client-of-bonita-bpm-engine.md
baptistemesta:
```suggestion
  In addition, starting from version 7.9.4, the number of connections used by the client can be configured with system property `org.bonitasoft.engine.api-type.connections.max`, the default is 20.
```

tbouffard:
Thanks @baptistemesta for the last changes
⚠️ I have just notice that the `two-main-types-of-deployment.md` page should also be updated: this new setting must be set otherwise this can lead to a communication bottleneck between the portal and the engine

baptistemesta:
The new setting's default value is 20. I don't think it will be a bottle neck if it's not set (compared to before ==2). But I can mention this property there.

educhastenier (approved):

tbouffard (approved):

baptistemesta:
BTW, commit title should be "feat" and not "fat" on merge ^^

tbouffard:
Release 7.9.4 publicly available so merging

PR #847, opened 2019-09-30
doc(deploy): Update title and desc

marielle-spiteri:
doc(deploy): Update title and desc

### DIFF
--- a/md/deploy-bundle.md
+++ b/md/deploy-bundle.md
@@ -1,6 +1,6 @@
 # Bonita Deploy bundle
 
-This page describes how to install Bonita platform on an existing and already installed Java EE Application Server like Tomcat or WildFly. 
+Description on how to install Bonita Runtime on an existing and already installed Java EE Application Server like Tomcat or WildFly. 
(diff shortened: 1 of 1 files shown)


md/deploy-bundle.md
```diff
@@ -1,4 +1,6 @@
-# Deploy bundle
+# Bonita Deploy bundle
+
+This page describes how to install Bonita platform on an existing and already installed Java EE Application Server like Tomcat or WildFly.
```
alachambre: Wildfly support has been removed on Bonita 7.10. 
So I suggest to just give Tomcat as example or to create a task to not forget to update this description in the 7.10 documentation in a few weeks, when changes have been propagated.

marielle-spiteri: Tomcat and WidFly are still supported to my knowledge, it must stay in the description to be precise. We should only remove it from 7.10 when it's released.

philippelaumay: agree with Marielle

PR #850, opened 2019-09-30
doc(comparison-of-7-x-and-6-x) : add description

Ilagouilly:
Adding a description to "Comparison of 7.x and 6.x" page.


md/comparison-of-7-x-and-6-x.md
```diff
@@ -1,5 +1,7 @@
 # Comparison of 7.x and 6.x
```
alachambre:
```suggestion
# Comparison of Bonita 7.x and Bonita 6.x
```

PR #962, opened 2019-10-03
doc(data-handling-overview): Update description and title

aibcmars:
Allow a better SEO

### DIFF
--- a/md/data-handling-overview.md
+++ b/md/data-handling-overview.md
@@ -1,5 +1,7 @@
 # Data handling overview
 
+How to effectively use the different kinds of data in the Bonita application
+
 In a BPM-based application, use:
 
 * [Business variables](define-and-deploy-the-bdm.md) for information that:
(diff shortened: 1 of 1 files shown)


md/data-handling-overview.md
```diff
@@ -1,5 +1,7 @@
 # Data handling overview
 
+How to effectively use the different kinds of data in the Bonita application
```
abirembaut: We usually talk about Bonita Platform in the rest of the documentation. So I would change the end of the sentence. What do you think of : "How to effectively use the different kinds of data available within Bonita Platform" ?

abirembaut (commented):
Just a comment on the description

md/data-handling-overview.md
aibcmars:
```suggestion
How to effectively use the different kinds of data available within Bonita Platform
```

aibcmars: I agree it should be consistent with other docs

PR #978, opened 2019-10-03
doc(pageName): Update description for page migration-overview

danila-m:
Allow a better SEO


md/migration-overview.md
```diff
@@ -1,4 +1,5 @@
 # Migration overview
+A overview of the migration process of a Bonita Platform.
```
educhastenier: 1. I think you should leave an empty line
2.
```suggestion
Learn how to migrate a Bonita Platform from an older version to a newer version.
```

educhastenier (changes requested):

PR #982, opened 2019-10-03
doc(create-your-first-project-with-the-engine-apis-and-maven): update description and title

educhastenier:


baptistemesta (approved):

abirembaut (changes requested):
It seems the title uses more than 50 characters.
I suggest: Create your first Application with the Engine API
and mention maven in the description

PR #1000, opened 2019-10-03
feat(subtasks): remove documentation on subtasks

baptistemesta:
this feature do not exists anymore starting from 7.3
I think it is safe to remove now

### DIFF
--- a/md/administrator-rights.md
+++ b/md/administrator-rights.md
@@ -9,6 +9,5 @@ A user with Administration rights has the maximum rights to manage activity on B
 * [Manage a process](processes.md)
 * [Manage a case](cases.md)
 * [Manage a task](tasks.md)
-* [Manage a subtask](subtasks.md)
 * [Manage a category](processes.md)
 * [Import/export an organization](organization-in-bonita-bpm-portal-overview.md)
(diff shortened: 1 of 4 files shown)


NathalieC:
ok to remove those. We need to make sure we still mention the fact that even if we don't provide the subtask creation feature after 7.3, we continue to execute subtasks if they have been created before 7.3.

baptistemesta:
I think it was the purpose of the page. We can keep it if we still need to mention that in 7.9

tbouffard:
Congrats @baptistemesta for PR number 1000
:clap:

educhastenier:
:question: so, do we remove the whole page or not ?

NathalieC:
So. It is ok to remove the page on subtasks. You are right. 
Here is the condition I'd like to see fulfilled: the note at the end of the "user-task-list" should say: 
"Since Bonita 7.3.0, creating subtasks is no longer possible. However, if subtasks have been created in a process instance in an old Bonita version, and if the Bonita Platform has been migrated to a later version while the case was open, then the subtasks will still be visible to the users assigned, and users will still be able to complete them."

educhastenier:
> So. It is ok to remove the page on subtasks. You are right.
> Here is the condition I'd like to see fulfilled: the note at the end of the "user-task-list" should say:
> "Since Bonita 7.3.0, creating subtasks is no longer possible. However, if subtasks have been created in a process instance in an old Bonita version, and if the Bonita Platform has been migrated to a later version while the case was open, then the subtasks will still be visible to the users assigned, and users will still be able to complete them."

I handle this !

PR #1063, opened 2019-10-14
doc(single-sign-on-with-cas): return URL addition

jjargot:
Host: header needs to be included in the incoming HTTP request to make the product forge the correct return URL, when the incoming HTTP request is not the same than the one originally used by  the user: there is a proxy in-between.

Versions: 7.6, 7.7, 7.8, 7.9


abirembaut:
Well done ! Excellent doc improvement

PR #1084, opened 2019-10-30
fix(themes) bootswatch version

rbioteau:
bootswatch version should be 3.3.7 (same as bootstrap)

### DIFF
--- a/md/themes.md
+++ b/md/themes.md
@@ -32,7 +32,7 @@ Export the examples to see how to structure a theme.
 
 Some of the default themes are based on the [bootswatch examples](https://bootswatch.com/), so you can easily choose another existing theme and package it to define a new theme.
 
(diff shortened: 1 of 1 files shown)


md/themes.md
```diff
@@ -35,7 +35,7 @@ Some of the default themes are based on the [bootswatch examples](https://bootsw
 A second way to create a custom theme is starting from an new custom theme in Studio and using a bootswatch default Sass theme. In order to do this using the version 3.3.4 of bootswatch:
```
DumitruCorini: 3.3.4 should be changed to 3.3.7 as well

DumitruCorini:
I tried the 3.3.7 and it seems to be the same as the 3.3.4 in terms of themes available, so it can be upgraded.

md/themes.md
rbioteau:
```suggestion
A second way to create a custom theme is starting from a new custom theme in Studio and using a Bootswatch default Sass theme. In order to do this using the version 3.3.7 of Bootswatch:
```

PR #1089, opened 2019-11-05
 feat(guest user): add documentation page for the guest user access

abirembaut:
Covers [BPO-265](https://bonitasoft.atlassian.net/browse/BPO-265)


md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+
+:::info 
+**Note:** For Enterprise, Performance, Efficiency, and Teamwork editions only.
+:::
+
+::: warning  
+ This feature is different from the auto-login feature that was available in the 6.x versions of Bonita BPM as it do gives directly access to a process form publicly. It is intended for living applications access.
+:::
+
+## Prerequisite: Guest profile and user account
```
passga: would prefer this title "Guest profile and dedicated user account"

passga (approved):
Perhaps it missing schema to explain more the behavior when a user like "walter.bates" is a member of public group and also of another group.

md/guest-user.md
abirembaut: done

md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Configure Bonita to allow guest user access to some applications
+
+The bundle already contains the files needed to setup guest user access with Bonita platform.
+To activate it:
+
+1.  Use the [platform setup tool](BonitaBPM_platform_setup) to retrieve the current configuration. You need to execute the following actions in the folder of the tenant in which the applications which requires public access are deployed.
+
+2. In the tenant_portal folder of the target tenant: `<BUNDLE_HOME>/setup/platform_conf/current/tenants/<TENANT_ID>/tenant_portal`,
+   edit the authenticationManager-config.properties as follows:
```
DumitruCorini: "Add the following lines to the *file*" instead of "edit the *file* as follows"? (or "Uncomment the following lines in the *file*" if they would be added after the beta as comments)

md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+    ```
+            [...]
+            # logout.link.hidden=true
+       -->  auth.tenant.guest.active=true
+       -->  auth.tenant.guest.username=guest
+       -->  auth.tenant.guest.password=guest
+       -->  auth.tenant.guest.apps=[public,guest] 
+    ```
+    
+    Make sure to set the username and password of the guest user account created in the section "Guest profile and dedicated user account".
```
DumitruCorini: An anchor towards the "Guest profile and dedicated user account" could be helpful since I couldn't understand where to find this section (since we are talking about the authenticationManager-config.properties file right now, I was thinking that it was in the authenticationManager-config.properties, but I guess you meant the section in the documentation page)

md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+:::info 
+**Note:** Since the applications will be public, it is recommended to have a group or a role containing all the users of Bonita platform organisation and add this group or role as a member of the profile used for public application (otherwise users already logged in with their account will get a 403 error when trying to access the public applications).
+:::
+
+## Configure Bonita to allow guest user access to some applications
+
+The bundle already contains the files needed to setup guest user access with Bonita platform.
+To activate it:
+
+1.  Use the [platform setup tool](BonitaBPM_platform_setup) to retrieve the current configuration. You need to execute the following actions in the folder of the tenant in which the applications which requires public access are deployed.
```
DumitruCorini: Link is fine, but I would still add an explanation on how to do it "by doing setup pull"? Or at least an an anchor on the setup pull so that the user doesn't need to read the entire page, but just the part that is important for this step.

md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+       -->  auth.tenant.guest.active=true
+       -->  auth.tenant.guest.username=guest
+       -->  auth.tenant.guest.password=guest
+       -->  auth.tenant.guest.apps=[public,guest] 
+    ```
+    
+    Make sure to set the username and password of the guest user account created in the section "Guest profile and dedicated user account".
+    The property "auth.tenant.guest.apps" contains the list of URL tokens of the applications that require public access (in this example, "public" and "guest"). Replace it with your applications tokens
+
+3. Use the [platform setup tool](BonitaBPM_platform_setup) again to save your changes into the database.
```
DumitruCorini: Same as for the setup pull

md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+       -->  auth.tenant.guest.username=guest
+       -->  auth.tenant.guest.password=guest
+       -->  auth.tenant.guest.apps=[public,guest] 
+    ```
+    
+    Make sure to set the username and password of the guest user account created in the section "Guest profile and dedicated user account".
+    The property "auth.tenant.guest.apps" contains the list of URL tokens of the applications that require public access (in this example, "public" and "guest"). Replace it with your applications tokens
+
+3. Use the [platform setup tool](BonitaBPM_platform_setup) again to save your changes into the database.  
+   Restart Bonita server.
```
DumitruCorini, line 53: I guess we can suppose that the user knows how to do this, but I'm not sure that he does.

DumitruCorini (changes requested):

md/guest-user.md
```diff
@@ -0,0 +1,59 @@
...
+::: warning  
+ This feature is different from the auto-login feature that was available in the 6.x versions of Bonita BPM as it do gives directly access to a process form publicly. It is intended for living applications access.
+:::
+
+## Guest profile and dedicated user account
+
++ In order to use this feature, you need to create a new user account in Bonita organisation. This account will be the one used to log in automatically a user that tries to access a public living application if he is not logged in to Bonita platform yet.
+
++ You also need to create a new custom profile and add the user account created in the previous step as a member of this profile. This profile will be the one to use as the profile of the living applications that require public access. If an application should also be accessible by users already logged in with their own bonita account, then they should also be members of this profile. 
+
```
alachambre, line 26: I don't understand something here:

If i'm correct, this `guest feature` works as follow: 
- A dedicated user is used as the guest user, this user is referenced in the configuration
- Any living application this user can access becomes accessible anonymously 

-> Why is it mandatory to create a custom profile? It doesn't work with the default profiles? 
for example, if my living app is associated to the profile User, and my anonymous account has the profile User, the app won't be accessible anonymously?

abirembaut: the app needs to be declared in the conf file below in order for the autologin as guest user to be performed when acessing it.
once you are logged in as guest, it is true that you can access any application this user can access.
It is not mandatory to create a custom profile but if you do not and use the user profile, it means you also guive access to the user profile in the portal to the guest user (just by modifying the URL). Which we don't want.

abirembaut: I added a warning to explain that

md/guest-user.md
abirembaut: done

md/guest-user.md
abirembaut: I added the information

md/guest-user.md
abirembaut: I added the information

md/guest-user.md
abirembaut: not here because it depends how bonita is deployed (all app servers are different...)

md/guest-user.md
abirembaut: I changed that
The lines will be in the file, you don't need to add them

md/guest-user.md
alachambre: 👍 
I guess that the portal doesn't need to be declared in the conf file to gain anonymous access because it is not a living app?

md/guest-user.md
DumitruCorini: Alright then

md/guest-user.md
abirembaut: giving access to the portal as guest is not something we want to enable. Only living apps

PR #1098, opened 2019-11-08
feat(docker): Add info about the docker deployment

danila-m:
Closes [BR-273](https://bonitasoft.atlassian.net/browse/BR-273)

### DIFF
--- a/md/two-main-types-of-deployment.md
+++ b/md/two-main-types-of-deployment.md
@@ -7,23 +7,18 @@ There are two main types of deployment
 using a Bonita bundle
 * Bonita Portal and Bonita Engine running on two different application servers
 
-## Foreword
-It is highly recommended to use the provided Tomcat bundle or the artifact `bonita.war` provided in the deploy bundle, in order to carry out these deployments successfully.
-
-There are two main types of deployment.
-
 ## Bonita Portal + Bonita Engine on the same application server
 
 ![deploy1](images/images-6_0/poss_deploy1.png)
 
-This is the simplest deployment configuration. The engine used is the one embedded in the webapp bonita.war. Using the pre-packaged Tomcat bundle is the easiest way to achieve this kind of deployment, but it is also possible to retrieve the `bonita.war` webapp provided in the **deploy.zip** and deploy it on another application server/servlet container.
-It is fast because the Bonita Portal and the Bonita Engine run on the same JVM and so there is no serialization and network overhead every time the Bonita Portal calls the engine.
+Bonita Portal and the Bonita Engine run on the same JVM. This is the simplest deployment configuration and is also the most performant as there is no serialization and network overhead.
+To perform this setup, you can use the [prepackaged bundle](tomcat-bundle.md) or the Bonita [docker image](bonita-docker-installation.md).
 
 **Advantages**
 
 * simple (single webapp and application server)
-* works out of the box if you use the provided Tomcat bundle
-* you can still access the embedded Bonita Engine API (or the Bonita Portal REST API) through HTTP if you need an external application to access it
+* works out of the box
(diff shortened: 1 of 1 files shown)


md/taxonomy.md
```diff
@@ -148,7 +148,7 @@
     * [Performance tuning](performance-tuning.md)
     * [Performance troubleshooting](performance-troubleshooting.md)
     * [Use gzip compression](use-gzip-compression.md)
-    * [Two main types of deployment](two-main-types-of-deployment.md)
+    * [Three main types of deployment](two-main-types-of-deployment.md)
```
baptistemesta: should we rename the page also?

md/two-main-types-of-deployment.md
```diff
@@ -29,6 +30,20 @@ It is fast because the Bonita Portal and the Bonita Engine run on the same JVM a
 **Drawbacks**
 
 * may not be adapted to some architecture constraints
+## Bonita Portal + Bonita Engine running in a docker container
+
+This is a new deployment option available in the 7.10 version.
```
baptistemesta:
```suggestion
This is a new deployment option available starting from 7.10 version.
```

baptistemesta: Also the docker image was available in other version. It's just that we document it now

md/two-main-types-of-deployment.md
```diff
@@ -29,6 +30,20 @@ It is fast because the Bonita Portal and the Bonita Engine run on the same JVM a
...
+**Advantages**
+
+* The usual advantages provided by using a docker container : isolation, simplicity, scalability...
+* Works out of the box with minimal effort.
+* Is the natural solution for a cluster or cloud deployment
+
+**Drawbacks**
+
+* Probably overkill for a simple on premise single server installation
+* More complicated installation than the previous solution
```
baptistemesta: I don't agree.
It's not overkill: it's more a matter of personal choice. If you already have some docker, it's easier to just run docker containers.
Also It's not more complicated installation (or it should not be). That's what you said in `Advantages`

IMHO, I would only put here something like `Require the installation of Docker`.

md/taxonomy.md
danila-m: We usually avoid it (dead links etc.)

md/two-main-types-of-deployment.md
```diff
@@ -29,6 +30,20 @@ It is fast because the Bonita Portal and the Bonita Engine run on the same JVM a
 **Drawbacks**
 
 * may not be adapted to some architecture constraints
+## Bonita Portal + Bonita Engine running in a docker container
+
+This is a new deployment option available starting from 7.10 version.
+The installation process is described [here](bonita-docker-installation.md)
+**Advantages**
```
educhastenier: New line for formatting: new paragraph, not on the same line
```suggestion

**Advantages**
```

md/two-main-types-of-deployment.md
educhastenier: Agreed with @baptistemesta. I would not put **any** of those drawbacks, as they are not, for me:
* It's not overkill because docker is thought to be the same on an on-premise installation and "in the cloud".
* It's not more complicated to run

educhastenier:
:question: General remark: is it a third main type of deployment, or is Docker an alternative that can be set up for both other types?
The question is open...

md/two-main-types-of-deployment.md
baptistemesta: maybe we can put as drawback that it could be harder to customize the installation... but I'm not sure

danila-m:
Reworked with Pablo.

pabloalonso:
LGTM

PR #1100, opened 2019-11-12
fix (portal-api): remove erroneous \ before _

marielle-spiteri:
remove erroneous \ before _ in search key and examples
Error starts in branch 7.4 up to 7.10 (it was fine in 7.3)


abirembaut:
LGTM

PR #1109, opened 2019-11-19
doc(DataManagement) Add documentation for data management

benjaminParisel:
Covers [UID-206](https://bonitasoft.atlassian.net/browse/UID-206)

### DIFF
--- a/md/ui-designer-overview.md
+++ b/md/ui-designer-overview.md
@@ -87,8 +87,6 @@ This synchronization is done by a new independent process and managed by Bonita 
 
 See [Data management](variables.md#business-data) section to know how to use this Business Data Model.
 
-If an error occurs, restart you Bonita Studio allow you restart this process.
-
 ## UI Designer target users' needed skills
 
 The UI Designer is intended for web developers creating pages and forms that will be part of a BPM application. It requires a good understanding of JavaScript and CSS.  
--- a/md/ui-designer-overview.md
+++ b/md/ui-designer-overview.md
@@ -80,6 +80,15 @@ In the Subscription editions, you can [update a form in production](live-update.
 
 To use a layout in a business application, export it from the Bonita UI designer, import it as a Bonita Portal resource, and map it as the layout of the relevant application.
 
+### Business Data model Synchronization
+
+Business Data Model is synchronized from Bonita Studio to Bonita UI-Designer when you deployed your BDM. 
+This synchronization is done by a new independent process and managed by Bonita Studio.
+
+See [Data management](variables.md#business-data) section to know how to use this Business Data Model.
+
+If an error occurs, restart you Bonita Studio allow you restart this process.
+
 ## UI Designer target users' needed skills
 
 The UI Designer is intended for web developers creating pages and forms that will be part of a BPM application. It requires a good understanding of JavaScript and CSS.  
--- a/md/variables.md
+++ b/md/variables.md
@@ -26,7 +26,8 @@ Expression, external API and URL parameter variables are evaluated every time so
 Business Data variables are meant to retrieve BDM objects defined in Studio. These objects are accessible through REST API calls, so they can be retrieved using External API variables defining the URL.
 For example, you can know by heart this URL:
 
-    `../API/bdm/businessData/{{businessDataType}}?q={{queryName}}&p=0&c=10&f={{filter}}`
+    ../API/bdm/businessData/{{businessDataType}}?q={{queryName}}&p=0&c=10&f={{filter}}
+    
 With: 
 *   **businessDataType** As example com.company.model.Invoice
 *   **queryName** One of object available queries, as example findByNumInvoice
@@ -40,9 +41,12 @@ This allows boost and simplify configuration on these data.
 A Business Data variable can be created in two ways:
 - From Data Model section in the left panel, drag and droping a Business Object to the whiteboard area
 - From the Variables bottom panel, creating a new variable with the type Business Data
+
 A configuration wizard appears: create a Business Data variable by providing information on how to retrieve elements of this Business Object.
+
 Select the query to retrieve Business Object. Queries can be a simple attribute query (ex: findByName) or an additional query such as a custom query.
 A filter area is displayed, depending on your selection, you may have to enter one or several attribute values. You can use a variable to fill its value using interpolation syntax.
+
 You can use Pagination section to define the number of elements you want by page (by default 10 elements) and the page number (by default first page - index 0).
 
 :::warning
--- a/md/variables.md
+++ b/md/variables.md
@@ -23,12 +23,30 @@ Expression, external API and URL parameter variables are evaluated every time so
 
 #### Business Data
 
-Business Data variable are used to fetch Business Data Model more easier than with external API. With this type. Just follow the wizard, select the query to retrieve Business Object, use dynamic variable (use `{{ myVariable }}`) or static value for filter parameter and set pagination data. Then you can bind your variable into widget or use it in Javascript Expression. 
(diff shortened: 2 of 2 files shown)


md/ui-designer-overview.md
```diff
@@ -80,6 +80,15 @@ In the Subscription editions, you can [update a form in production](live-update.
 
 To use a layout in a business application, export it from the Bonita UI designer, import it as a Bonita Portal resource, and map it as the layout of the relevant application.
 
+### Business Data model Synchronization
```
jeromecambon, line 83:
```suggestion
### Business Data Model (BDM) and UI Designer
```

md/ui-designer-overview.md
```diff
@@ -80,6 +80,15 @@ In the Subscription editions, you can [update a form in production](live-update.
...
+
+Business Data Model is synchronized from Bonita Studio to Bonita UI-Designer when you deployed your BDM.
```
jeromecambon, line 85:
```suggestion
The Business Data Model, defined in Bonita Studio, is exposed in the UI Designer.
```

md/ui-designer-overview.md
```diff
@@ -80,6 +80,15 @@ In the Subscription editions, you can [update a form in production](live-update.
...
 To use a layout in a business application, export it from the Bonita UI designer, import it as a Bonita Portal resource, and map it as the layout of the relevant application.
 
+### Business Data model Synchronization
+
+Business Data Model is synchronized from Bonita Studio to Bonita UI-Designer when you deployed your BDM. 
+This synchronization is done by a new independent process and managed by Bonita Studio.
+
+See [Data management](variables.md#business-data) section to know how to use this Business Data Model.
+
+If an error occurs, restart you Bonita Studio allow you restart this process.
```
jeromecambon: Finally, I think we should say this here. It is more a trouble shooting stuff, that should not appear in the general section.
Not sure we have a trouble shooting section, so the icon+tooltip in the UID should be enough.

md/ui-designer-overview.md
```diff
@@ -80,6 +80,15 @@ In the Subscription editions, you can [update a form in production](live-update.
 
 To use a layout in a business application, export it from the Bonita UI designer, import it as a Bonita Portal resource, and map it as the layout of the relevant application.
 
+### Business Data model Synchronization
+
+Business Data Model is synchronized from Bonita Studio to Bonita UI-Designer when you deployed your BDM. 
+This synchronization is done by a new independent process and managed by Bonita Studio.
```
jeromecambon, line 86:
```suggestion
The synchronization is done by a new independent process and managed by Bonita Studio.
```

md/variables.md
```diff
@@ -12,13 +12,29 @@ There are several types of variables:
...
 
 ## Evaluation cycle
 
 String and JSON variables are initialized at page data model instantiation and their values do not change subsequently. 
 
 Expression, external API and URL parameter variables are evaluated every time something happens on the page, such as when receiving data or on a user interaction. The call of an External API variable is triggered every time the URL of the resource it points at changes.
 
+#### Business Data
+
+Business Data variable are used to fetch Business Data Model more easier than with external API. With this type. Just follow the wizard, select the query to retrieve Business Object, use dynamic variable (use `{{ myVariable }}`) or static value for filter parameter and set pagination data. Then you can bind your variable into widget or use it in Javascript Expression.
```
jeromecambon:
```suggestion
Business Data variables are used to fetch Bonita Business Data easier than with Business Data API.

A Business Data variable can be created in two ways:
- From Data Model section in the left panel, drag and droping a Business Object to the whiteboard area
- From the Variables bottom panel, creating a new variable with the type Business Data

A configuration wizard appears: create a Business Data variable by providing information on how to retrieve elements of this Business Object.

==> Copy <Define how to retrieve Business Object and Use your Business data> from embedded doc here.

```

md/variables.md
```diff
@@ -12,13 +12,29 @@ There are several types of variables:
...
 
 String and JSON variables are initialized at page data model instantiation and their values do not change subsequently. 
 
 Expression, external API and URL parameter variables are evaluated every time something happens on the page, such as when receiving data or on a user interaction. The call of an External API variable is triggered every time the URL of the resource it points at changes.
 
+#### Business Data
+
+Business Data variable are used to fetch Business Data Model more easier than with external API. With this type. Just follow the wizard, select the query to retrieve Business Object, use dynamic variable (use `{{ myVariable }}`) or static value for filter parameter and set pagination data. Then you can bind your variable into widget or use it in Javascript Expression. 
+
+You can fire this wizard with drag and drop a Object from Data Model section in left palette panel to white-board area. This wizard will only create a new variable and it can be used in your page.
```
jeromecambon: To remove (see above)

md/variables.md
```diff
@@ -12,13 +12,29 @@ There are several types of variables:
...
+
+:::warning
+Be careful, by default each business data variable is an array (=collection), even if only one result from query is returned.
```
jeromecambon:
```suggestion
Be careful, by default each business data variable is an array (collection), even if only a single item is returned from query.
```

md/variables.md
```diff
@@ -12,13 +12,29 @@ There are several types of variables:
...
+:::
+
+:::info
+Read [Integration with Bonita Platform](ui-designer-overview.md#integration-with-bonita-platform) to know how Business Data are connected with Bonita Platform.
```
jeromecambon, line 57: ?? Not sure I understand this reference

jeromecambon (changes requested):

md/variables.md
pabloalonso: "Business Data variables are used to fetch Bonita Business Data easier than with Business Data API."
...business business business

md/variables.md
pabloalonso: not sure how to phrase this: "...even if only a single item is returned from query"

I think we should be more positive, meaning:
"Business data variables will be always an array that contains the result of query execution. In case of single result queries, such as *find by persistence Id*, you may access the values like this *InvoiceVar[0]*, knowing that 0 is first and only element of your array."

md/variables.md
pabloalonso: Business Data variables are meant to retrieve BDM objects defined in Studio. These objects are accesible through REST API calls, so they were retrieved using External API variables defining the URL. 
For example, using this URL: 
```
../API/bdm/businessData/{{businessDataType}}?q={{queryName}}&p=0&c=10&f={{filter}}
```
It implied to know by heart:
 ``businessDataType`` As example *com.company.model.Invoice*
 ``queryName`` One of object available queries, as example *findByNumInvoice*
 ``filter`` All filters required by the query, as example *f=numInvoice=XXXX*

With Business Data variables, you are able to select one object (from left panel or select box), see its available queries and select one, and configure its filters which are already proposed.
This allows boost and simplify configuration on these data

jeromecambon: Interesting details for a Fred that were using previous UID versions.
But for a Cindy that just start using UID, it can be confusing...
I suggest to put this in a dedicated section, such as "Details for users that used previous UI Designer versions"

jeromecambon: > Business Data variables are meant to retrieve BDM objects defined in Studio. These objects are accesible through REST API calls, so they were retrieved using External API variables defining the URL.

Business Data variables are meant to retrieve BDM objects defined in Studio. These objects are accessible through REST API calls, so they can be retrieved using External API variables defining the URL.

> With Business Data variables, you are able to select one object (from left panel or select box), see its available queries and select one, and configure its filters which are already proposed.
This allows boost and simplify configuration on these data

I would just say that it is not needed anymore with Business Data variables.

pabloalonso (approved):
LGTM

PR #1112, opened 2019-11-27
getting-started: fix screenshots

amottier:
A very simple screenshot update that display expected result after doing instruction described in step 6 right before the screenshot display on https://documentation.bonitasoft.com/bonita/7.9/define-who-can-do-what


baptistemesta:
Is it only from 7.9?

amottier:
Yes this update start with version 7.9. It will apply also on newer version.

PR #1115, opened 2019-12-02
doc(UID-166_table_key_number): Added limitation warning for Table and Data Table widgets

jeromecambon:
- Limitation is about 'column keys' property of Table and Data Table

Covers [UID-166](https://bonitasoft.atlassian.net/browse/UID-166)

### DIFF
--- a/md/widgets.md
+++ b/md/widgets.md
@@ -251,7 +251,8 @@ Use a display widget for information that the user can read on a page. This incl
 
 Use a table widget to display data in a table. In order to display the data, first define the **headers** property with a comma-separated list of column headings.  
 Then bind the **content** to an array of JavaScript objects.  
-Finally, provide a comma-separated list for the **columns keys** indicating the attribute to display in each column.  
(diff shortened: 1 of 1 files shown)


amottier:
@jeromecambon "These column keys must not be a number, or start with a number" => if it's a number it start with a number so I guess we can simplify the message to "These column keys must not start with a number". Same apply to https://github.com/bonitasoft/bonita-ui-designer-sp/pull/2838 but as it's minor probably don't worth the effort.

benjaminParisel (approved):
LGTM

jeromecambon:
> @jeromecambon "These column keys must not be a number, or start with a number" => if it's a number it start with a number so I guess we can simplify the message to "These column keys must not start with a number". Same apply to [bonitasoft/bonita-ui-designer-sp#2838](https://github.com/bonitasoft/bonita-ui-designer-sp/pull/2838) but as it's minor probably don't worth the effort.

Thanks for your comment. I think it is better to be explicit, to avoid confusion. Discussing with Pablo & Benjamin, they agree to keep like this.

PR #1118, opened 2019-12-06
fix(bundle): path to request_key_utils

tbouffard:
The tool version has been removed from the path in GA but the doc still mention it


educhastenier (approved):
Nice Catch :ok_hand:

PR #1122, opened 2020-01-13
fix(REST API) document delete process with warning

abirembaut:
- document delete process with REST API
- add data loss warning

Covers [BPO-330](https://bonitasoft.atlassian.net/browse/BPO-330)


julienmege (commented):
LGTM

PR #1124, opened 2020-01-15
doc(SAML) Hash URL parameters lost in redirections

abirembaut:
-document limitation and workaround

Covers [BPO-335](https://bonitasoft.atlassian.net/browse/BPO-335)


julienmege (commented):
The bookmark won't be directly usable in the GWT Portal,
but LGTM,
As we cannot do better than this...

PR #1126, opened 2020-01-17
doc (wildfly): Fix typos remove useless instr

marielle-spiteri:
I don't think that the following step is necessary, especially if the user aren't using the bonita bundle. Unless it actually breaks the tomcat install? Deleting the old wildfly isn't a useful step to mention IMO: 
* (delete your old WildFly installation by removing the Bonita WildFly root folder)
We don't support wildFly, period, we don't need to say Wildfly bundle. 
+ Fix several typos.

### DIFF
--- a/md/convert-wildfly-into-tomcat.md
+++ b/md/convert-wildfly-into-tomcat.md
@@ -4,7 +4,7 @@ Learn how to convert a Bonita installation based on WildFly to a one based on To
 
 ## Overview
 
-Bonita 7.10 does not provide nor support any WildFly anymore. As a consequence, it is necessary when updating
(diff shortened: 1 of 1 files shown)


md/convert-wildfly-into-tomcat.md
```diff
@@ -4,22 +4,21 @@ Learn how to convert a Bonita installation based on WildFly to a one based on To
 
 ## Overview
 
-Bonita 7.10 does not provide nor support any WildFly bundle anymore. As a consequence, it is necessary when updating
+Bonita 7.10 does not provide nor support any WildFly anymore. As a consequence, it is necessary when updating
```
educhastenier:
```suggestion
Bonita 7.10 does not provide nor support WildFly anymore. As a consequence, it is necessary when updating
```

PR #1127, opened 2020-01-21
doc(CAS) Add note regarding Guest user access

abirembaut:
- add note regarding the version of CAS server it was tested with

Covers [BPO-376](https://bonitasoft.atlassian.net/browse/BPO-376)


julienmege (commented):
LGTM

PR #1129, opened 2020-01-22
Fix issue with display of large images

amottier:
In order to avoid horizontal scroll, add  after images included using markdown syntax (same solution as seen on user-task-list page).


educhastenier (approved):
Didn't check but final display, but looks good to me :+1:

educhastenier (approved):

PR #1131, opened 2020-01-23
fix(Licenses): Improve heading for licence replacement

ugaston:
Versions: 7.7, 7.8, 7.9, 7.10

Choose a better wording + add condition for the delete of all old licences.


jjargot (approved):
perfect !

PR #1135, opened 2020-01-27
fix (internal) SQL server to support NT Authentication

marielle-spiteri:
Adding example with integratedSecurity=true for SQL server to support NT Authentication


educhastenier (commented):
We have to discuss about this PR because I don't understand the context

PR #1147, opened 2020-02-03
fix(oracle-driver) clarify oracle drivers usage

marielle-spiteri:
- Fix the URLs to oracle drivers.
- Clarify versions and add Checksum for each driver.

Versions: 7.6, 7.7, 7.8, 7.9, 7.10


jjargot (approved):
Thx Marielle

PR #1148, opened 2020-02-03
fix(oracle-driver) Clarify driver version to use

marielle-spiteri:
- Add warning
- Add checksum check for the official driver

Versions: 7.9, 7.10


jjargot (approved):
Thx Marielle

PR #1149, opened 2020-02-04
fix(oracle-drivers-check) Warning about drivers

marielle-spiteri:
- add a warning to use correct oracle driver


jjargot (approved):
Thx Marielle

PR #1150, opened 2020-02-04
fix(oracle-drivers) Explicit drivers version

marielle-spiteri:
- Extra warning for oracle drivers version in RDBMS requirements
- Adding a warning for oracle driver in lib folder

Versions: 7.6, 7.7, 7.8, 7.9, 7.10


jjargot (approved):
thx MArielle

PR #1152, opened 2020-02-05
fix(migration) Steps numbering fix

marielle-spiteri:
- Fix the migration full procedure numbering, based on https://stackoverflow.com/questions/18088955/markdown-continue-numbered-list


educhastenier (approved):
Looks good to me :+1:
Thx for this contribution

PR #1178, opened 2020-03-03
fix(Custom Deployment): add warning about tomcat libraries

ugaston:
Versions: 7.9, 7.10

Add a warning about the fact that there may be a discrepancy between the tomcat libraries copied from the bundle's lib/bonita and the target Tomcat's lib.


jjargot (approved):
thx Unai

PR #1210, opened 2020-04-23
feat(CORS) improve HTML Example test page

julienmege:
- use fetch instead of jquery
- Allows to edit user password without hard-coded username and password

cover [BPO-540](https://bonitasoft.atlassian.net/browse/BPO-540)


nchabanoles (approved):
This is great, I would have 2 suggestions if you want to go further:
* define a form input for the BONITA_ACCESS_URL to avoid the user to modify scripts (limit mistakes)
* log errors in a div instead of the console, it makes it easier to test without to open the dev tools
other than that LGTM

abirembaut:
LGTM

PR #1213, opened 2020-04-28
fix(SAML SSO): document whitelist option

ugaston:
Versions: 7.10

The whitelist option had only been documented in Kerberos chapter, but it works with SAML too.


abirembaut:
LGTM
Good job ! :+1:

PR #1216, opened 2020-05-04
doc(release notes): portal transformation info

NathalieC:
* add portal transformation info

### DIFF
--- a/md/release-notes.md
+++ b/md/release-notes.md
@@ -83,6 +83,18 @@ Aside of the quality of life update of not having to run a migration between mig
 * A check has been added in Cluster mode as to forbid the start of nodes in different Bonita platform versions. For example, on a two-node cluster, you can't have a node in 7.11.0 and another in 7.11.1.
 The feature of having nodes in different Bonita versions was never supported, though there were cases where it "worked" in previous versions.  
 
+## Bonita Portal transformation
+
+Bonita Portal is being transformed into Bonita Applications. When Bonita Applications are ready, Bonita Portal will be removed.
+Developers and users will need to learn how to stop using the Portal and start using Bonita Applications instead.
+This change will allow Bonita and its users to get free from Google Web Toolkit (GWT) technology and offer opportunities for customization.
+Indeed, some Portal pages (with GWT inside) are being totally recreated with our own UI Designer. They will be customizable.
+Others (those that already were GWT-free) are being wrapped and will not be customizable.
+Moreover, as any Living Application, Bonita applications will be extensible to add any page the users need.
+More details in the upcoming versions of Bonita.
+Until then, we strongly advise not to create Custom Portal Profiles anymore but applications instead if possible.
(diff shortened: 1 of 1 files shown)


md/release-notes.md
```diff
@@ -83,6 +83,18 @@ Aside of the quality of life update of not having to run a migration between mig
 * A check has been added in Cluster mode as to forbid the start of nodes in different Bonita platform versions. For example, on a two-node cluster, you can't have a node in 7.11.0 and another in 7.11.1.
 The feature of having nodes in different Bonita versions was never supported, though there were cases where it "worked" in previous versions.  
 
+## Bonita Portal transformation
+
+Bonita Portal is being transformed into Bonita Applications. When Bonita Applications are ready, Bonita Portal will be removed.
+Developers and users will need to learn how to stop using the Portal and start using Bonita Applications instead.
+This change will allow Bonita and its users to get free from Google Web Toolkit (GWT) technology and offer opportunities for customization.
+Indeed, some Portal pages (with GWT inside) are being totally recreated with our own UI Designer. They will be customizable.
```
abirembaut: I would say "(built with GWT)" instead of "(with GWT inside)"

md/release-notes.md
```diff
@@ -83,6 +83,18 @@ Aside of the quality of life update of not having to run a migration between mig
...
+Others (those that already were GWT-free) are being wrapped and will not be customizable.
```
abirembaut: (those that were already using another technology) instead of (those that already were GWT-free)

md/release-notes.md
```diff
@@ -83,6 +83,18 @@ Aside of the quality of life update of not having to run a migration between mig
...
+Moreover, as any Living Application, Bonita applications will be extensible to add any page the users need.
+More details in the upcoming versions of Bonita.
+Until then, we strongly advise not to create Custom Portal Profiles anymore but applications instead if possible.
+When Bonita Portal does not exist anymore, the existing Portal Custom Profiles will be migrated into Living applications.
```
abirembaut: "When Bonita Portal **won't** exist anymore" maybe
"will **need to** be migrated into Living applications" maybe because we don't know yet if it will be automatic or if we will provide a tool or whaterver

abirembaut (changes requested):
just a few comments.
Regarding GWT, it is used to build the page and it is a tool more than an something used during execution. hence my comments

md/release-notes.md
NathalieC: fixed. Thanks.

md/release-notes.md
NathalieC: fixed thanks

md/release-notes.md
NathalieC: We cannot say When + will in english. 
Fixed "need to", you are totally right. Thanks.

abirembaut: :+1:

abirembaut (approved):

PR #1238, opened 2020-05-26
fix(ldap-sync): uniformize table formatting

pabloalonso:
Use same formatting as below, a table

### DIFF
--- a/md/ldap-synchronizer.md
+++ b/md/ldap-synchronizer.md
@@ -318,14 +318,12 @@ Groups will be synchronized based on the matching of their LDAP common name (CN)
 
 
 Groups can be declared individually in the configuration file with the following properties :
-ldap\_group\_dn
-mandatory attribute that specifies the DN of the LDAP group.
 
-forced\_bonita\_group\_name
-optional attribute that renames the Bonita group instead of using the original LDAP group name.
-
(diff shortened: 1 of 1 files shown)


md/ldap-synchronizer.md
```diff
@@ -353,7 +351,7 @@ In combination or as an alternative, groups can be declared using the result of
 |:-----|:--------|
 | ldap\_group\_search\_dn | DN of the LDAP watched directory that will be used to get the list of the LDAP groups. |
 | ldap\_group\_search\_filter | LDAP group search filter (mandatory attribute, but can be a wide filter such as "cn=\*"). |
-| force\_add\_non\_existing\_users | optional boolean attribute (true by default) that defines whether group members that are not present in Bonita should be imported (if false, these users are ignored). |
+| force\_add\_non\_existing\_users | optional boolean attribute (true by default) that defines whether group members (users in LDAP) that are not present in Bonita should be imported (if false, these users are ignored but group is created). |
```
danila-m:
```suggestion
| force\_add\_non\_existing\_users | optional boolean attribute (true by default) that defines whether group members (users in LDAP) that are not present in Bonita should be imported (if false, these users are ignored but group is created/membership changes are still applied). |
```

PR #1233, opened 2020-05-19
feat(cors) Choose cookies SameSite Policy

julienmege:
covers [BPO-540](https://bonitasoft.atlassian.net/browse/BPO-540)

PR #1240, opened 2020-05-28
fix : Tomcat configuration

NithoPamc:
Correction of the Tomcat configuration file because the following error appeared in the browser :

https://developer.mozilla.org/fr/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials?utm_source=devtools&utm_medium=firefox-cors-errors&utm_campaign=default

I only have version 7.8, but the error should be on the following versions.

In my case, bonita was accessible at this address: http://127.0.0.1:8080/bonita/
And my page was hosted here: http://127.0.0.1/

### DIFF
--- a/md/enable-cors-in-tomcat-bundle.md
+++ b/md/enable-cors-in-tomcat-bundle.md
@@ -23,7 +23,7 @@ Edit the web.xml of the bonita.war to add the CORS filter:
   <filter-class>org.apache.catalina.filters.CorsFilter</filter-class>
   <init-param>
     <param-name>cors.allowed.origins</param-name>
-    <param-value>*</param-value>
+    <param-value>[DOMAIN_OF_YOU_HOSTED_PAGE]</param-value>
   </init-param>
 
   <init-param>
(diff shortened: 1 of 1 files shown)


md/enable-cors-in-tomcat-bundle.md
```diff
@@ -58,6 +64,8 @@ _**Important Note 1:** The filter must be inserted in the file webapps/bonita/WE
 
 _**Important Note 2:** It must be the first filter, inserted right after the </error-page> tag_
 
+_**Important Not 3:** Replace [DOMAIN_OF_YOU_HOSTED_PAGE] with the domain of your web page (Eg: http://mydomaine.com)_
```
educhastenier, line 67:
```suggestion
_**Important Note 3:** Replace [DOMAIN_OF_YOU_HOSTED_PAGE] with the domain of your web page (Eg: http://mydomain.com)_
```

DumitruCorini:
This proposition is good, but we already fixed this for 7.11 in [this pr](https://github.com/bonitasoft/bonita-doc/pull/1233). This will be reported into all other versions shortly.

PR #1242, opened 2020-06-03
32bit installer issues with  macos10.15

florentsarat:


alachambre:
retest this please

alachambre:
I suggest to not number the two points , it doesn't render well with the danger area (see screenshot)

Capture d’écran 2020-06-08 à 09 43 02

PR #1252, opened 2020-06-22
update(index): add archetype, fix uiPath

NathalieC:
* add archetype link to release note as an important value
* fix non-working link to UiPath Cloud connector


alachambre (approved):
:)

PR #1253, opened 2020-06-22
update (release notes): add IE11 support deprec

NathalieC:
* add IE11 support deprecation
* add anchor for uipath cloud, add anchor for connector archetype so it can be reached from index.md
* fix link in info well to portal transformation


benjaminParisel (approved):
LGTM

PR #1263, opened 2020-07-24
fix(enforce-password-policy.md) : updating the link to security-config.properties

Ilagouilly:
Updating the link to the platform setup tool instead of security-config.properties.


Ilagouilly:
Additionally, the location of the file could be updated, but I wanted to check this with someone before.

<BUNDLE_HOME>/setup/platform_conf/initial/platform_portal/security-config.properties
<BUNDLE_HOME>/setup/platform_conf/current/platform_portal/security-config.properties

educhastenier:
@Ilagouilly please rebase your PR, as it contains 550 commits for now!

To answer your question:
* initial folder is the place to modify **if you did not start the server for the first time yet**
* current folder is the place to modify **otherwise**

Ilagouilly:
@educhastenier I knew something was off when I was pushing this. I'll fix this asap ;)
Also thanks for answering my question, but it was more about if whether we should include the location of the security-config.properties file (initial or current) in the doc page too or not :)

PR #1266, opened 2020-07-29
doc(dependencies) Initiate dependencies index page

benjaminParisel:


benjaminParisel:
See with @NathalieC and it's ok

PR #1284, opened 2020-10-02
feat(dev-doc): add developer doc flow node stats

baptistemesta:
There are existing pages about this subject that can be linked there or
removed

### DIFF
--- a/md/engine-flow-node-states.md
+++ b/md/engine-flow-node-states.md
@@ -1,0 +1,88 @@
+# Flownode States
+
+Each flownode has a Type. The flownode type determines which states the flownode can be in, and the sequence of the states it will follow during its lifecycle.
+Each flownode object has a state (_FlowNodeState_) at a given time. The state of the Flownode determines how it will be handled by the Bonita Engine.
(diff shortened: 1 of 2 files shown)


md/engine-flow-node-states.md
```diff
@@ -0,0 +1,88 @@
...
+
+If the state is finished, the State Machine determines the next state:  
+Each flownode type has an ordered list of states.  
+We take the next state in the list, and ask the state to determine whether we should execute the state or not.  
+If not, the next state is "skipped" and the second next state is checked the same way, until a state determines it should be executed.  
+If so, the next state is executed, in another background task (if terminal or not stable).  
+
+
+## Example
+A UserTask is...
```
educhastenier: incomplete.

educhastenier:
> There are existing pages about this subject that can be linked there or
> removed

So, what do we decide? Should we reference those pages? Should we remove / merge them?

md/engine-flow-node-states.md
baptistemesta: removed it

PR #1313, opened 2020-10-21
doc(login) redirect parameter is now optional in 7.12

abirembaut:
- If there is a redirect param in the request it is used. Otherwise 
check if there is a redirect URL
- update release note

This means it is no longer needed to put redirect=false in the request
to log in using the API. But previous login requests with a redirect URL
will continue working as the redirect parameter is optional

Covers [BPO-655](https://bonitasoft.atlassian.net/browse/BPO-655)

To be merged with bonitasoft/bonita-web-sp/pull/622

### DIFF
--- a/md/rest-api-authentication.md
+++ b/md/rest-api-authentication.md
@@ -9,7 +9,7 @@ To log in, use the following request:
 | Request URL | `http://host:port/bonita/loginservice`|
 | Request Method | POST|
 | Content-Type | application/x-www-form-urlencoded|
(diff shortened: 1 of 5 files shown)


md/rest-api-authentication.md
```diff
@@ -57,9 +57,9 @@ To log out, use the following request:
 |:-|:-|
 | Request URL | `http://host:port/bonita/logoutservice`|
 | Request Method | GET|
-| Query parameter | redirect: true or false (default set to true)|
+| Query parameter | redirect: `true` or `false` (set to `false` by default)|
 
-Setting the redirect parameter to false indicates that the service should not redirect to the login page after logging out.
+Not setting then redirect parameter to `true` means that the service will not redirect to the login page after logging out.
```
baptistemesta:
```suggestion
Not setting the redirect parameter to `true` means that the service will not redirect to the login page after logging out.
```

PR #1340, opened 2020-11-18
Add(ui-designer-troubleshooting): images

marielle-spiteri:
* Adding images for ui-designer-troubleshooting page


jeromecambon:
Added the images in the UID Troubleshooting page

PR #1344, opened 2020-11-25
doc(events): add constraint to event handlers implementation

baptistemesta:


md/event-handlers.md
```diff
@@ -6,8 +6,8 @@ An event handler is an extension to the engine that is configured to run when a
 **Note:** For Enterprise, Performance, Efficiency, and Teamwork editions only.
 :::
 
-You can add event handlers for several purposes and you can configure which events you want to catch.
-We strongly recommend that you add only appropriate handlers and carefully code the handler filters to handle only those events that you are interested in.
+Event handlers can help in adding new features to the Platform. You can use it to call your own services when specific events
+happens like a case starts, or a user submit a task.
```
educhastenier:
```suggestion
Event handlers can help in adding new features to the Platform. You can use them to call your own services when specific events
happen like when a case starts, or when a user submits a task.
```

md/event-handlers.md
```diff
@@ -17,6 +17,16 @@ You can create an event handler to track any change to any object in the databas
...
 
+
+## Constraints
+
+The platform executes events handlers synchronously at the exact moment the related entity is updated, therefore it executes 
+it in the same thread and same transaction that the one executing the element.
+
+It means that:
+* Events handlers should never block, i.e. no network call, no I/O on the filesystem.
+* Events handlers should never throw exception, if it does, the related element execution will fail.
```
educhastenier:
```suggestion
* Events handlers should never throw exception, if it does, the related element execution will fail, meaning that the transaction is rolled back and the task (for instance) goes to the failed state, which is generally NOT what one wants.
```

educhastenier (commented):
🧲  Maybe also add details regarding what cannot / should not be called from inside an Event Handler, like calling APIs, ...

baptistemesta:
> 🧲 Maybe also add details regarding what cannot / should not be called from inside an Event Handler, like calling APIs, ...

Nice idea. I will do that.

educhastenier:
> > magnet Maybe also add details regarding what cannot / should not be called from inside an Event Handler, like calling APIs, ...
> 
> Nice idea. I will do that.

Ok, tell me when it's done

PR #1345, opened 2020-11-25
doc(recovery): add more info on recovery metrics and conf

baptistemesta:


md/fault-tolerance.md
```diff
@@ -31,9 +31,12 @@ Configuration for the retry is available in the `bonita-tenant-comunity-custom.p
 [setup tool](BonitaBPM_platform_setup.md)
 
 ```properties
-# number of times the works will be retried in case of issues that can be retried
+# Retry mechanism: retry work when they fails because of an error that is transient
```
educhastenier:
```suggestion
# Retry mechanism: retry the works when they fail because of an error that is transient
```

md/fault-tolerance.md
```diff
@@ -116,3 +129,7 @@ bonita_bpmengine_recovery_recovered_total_elements_total{tenant="1",} 39768.0
 # TYPE bonita_bpmengine_recovery_execution_executions_total counter
 bonita_bpmengine_recovery_execution_executions_total{tenant="1",} 818.0
 ```
+
+
+### Troubleshooting
```
educhastenier: Is this section added empty?

md/fault-tolerance.md
```diff
@@ -59,16 +62,21 @@ Also, in cluster environment, only one node is responsible to run the recovery a
...
+# Recovery Mechanism: recreate works when they are lost due to incidents
+# All following configuration should work for everyone, it can be changed only to do performance tuning in limit-cases
 # Avoid verifying elements recently modified, by default no elements updated during the last hour is considered (ISO-8601 duration format).
 bonita.tenant.recover.consider_elements_older_than=PT1H
 # Duration after the end of the previous execution before a new one is started. By default recovery runs every 2 hours (ISO-8601 duration format)
 bonita.tenant.recover.delay_between_recovery=PT2H
-
 ```
 
+`bonita.tenant.recover.delay_between_recovery` is also the delay before the first recovery after the startup.
```
educhastenier: Does it mean the recovery does not start right away at platform startup?
2 hours can seem a lot then !

md/fault-tolerance.md
```diff
@@ -116,3 +129,7 @@ bonita_bpmengine_recovery_recovered_total_elements_total{tenant="1",} 39768.0
 # TYPE bonita_bpmengine_recovery_execution_executions_total counter
 bonita_bpmengine_recovery_execution_executions_total{tenant="1",} 818.0
 ```
+
+
+### Troubleshooting
+
```
baptistemesta:
```suggestion
```

md/fault-tolerance.md
baptistemesta: We still restart elements at startup. It is not considered as the recovery mechanism.

As explained here: the recovery mechanism `recreate works when they are lost due to incidents`

That means we wait 2 hours before checking for works lost due to an incident that happened since the startup.

baptistemesta: To be more clear: 
* We restart all elements at startup, regardless of their "age"
* 2 hours later and then every 2 hours, we run the recovery mechanism that recover elements only "older" than 1 hour

educhastenier: > To be more clear:
> 
> * We restart all elements at startup, regardless of their "age"
> * 2 hours later and then every 2 hours, we run the recovery mechanism that recover elements only "older" than 1 hour

This part is much clearer that way. Can we include it in the doc like that? 👍🏼

baptistemesta: I've updated that is the description of the recovery mechanism

PR #1385, opened 2021-01-22
update(migration): services pause and bdm warnings

marielle-spiteri:
* change sentences and warning regarding the BDM update and pausing Services


educhastenier:
@pabloalonso Can you share with @mariellebonita the action plan we define regarding the BPM pause strategy for migration, as I think it is incompatible with this PR?

baptistemesta:
Merging that before the doc migration, It needs to be re-vamped because of the migration bug in any case

PR #1392, opened 2021-01-28
doc(externalApi): Add status and headers properties

benjaminParisel:
* Explain how to get response status code and response header from an external API call
* Available in 2020.2 version


jeromecambon (approved):
LGTM

PR #1394, opened 2021-01-28
Update(doc): wrong information removal

marielle-spiteri:
* This was wrong, as the tool will raise an error if either or both mandatory parameters are missing: # If no PROCESS_DEFINITION_ID parameter is given, the program lists all existing process definitions and stops: bin/bonita-purge-tool
 * Adding the mention mandatory in the sentence explaining how to use the parameters that are mandatory 
* To be pushed in doc 7.9  and higher

### DIFF
--- a/md/purge-tool.md
+++ b/md/purge-tool.md
@@ -24,26 +24,20 @@ Enter your database configuration properties in the file `application.properties
 
 ## Run Bonita Purge Tool
 
-This command will delete all archived process instances belonging to the process identified by **PROCESS_DEFINITION_ID**,
+This command will delete all archived process instances belonging to the process identified by mandatory parameters **PROCESS_DEFINITION_ID**,
 that are finished since at least **OLDEST_DATE_TIMESTAMP**.
 
 
 example (Unix):
 
 ```shell
-# If no PROCESS_DEFINITION_ID parameter is given, the program lists all existing process definitions and stops:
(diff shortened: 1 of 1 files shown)


educhastenier:
> * This was wrong, as the tool will raise an error if either or both mandatory parameters are missing: # If no PROCESS_DEFINITION_ID parameter is given, the program lists all existing process definitions and stops: bin/bonita-purge-tool
> * Adding the mention mandatory in the sentence explaining how to use the parameters that are mandatory
> * To be pushed in doc 7.9  and higher

Thx @mariellebonita for noticing this.
The tool however should behave as described: when processDefinitionId is missing, it should list all the available processes and stop.
If it doesn't, this means there is a bug / regression. Can you create a bug in Jira for that, please?

Putting this PR in 🚧  to engage discussion.

marielle-spiteri:
Hi @educhastenier , I have created the bug https://bonitasoft.atlassian.net/browse/RUNTIME-130

educhastenier:
Ok @mariellebonita, so as the bug is being fixed, I suggest you close this PR, right?
Then regarding when to release a new purge tool version, I let the competent decisioners handle that 😜

marielle-spiteri:
@educhastenier , sounds good, closing the PR, but hopefully the Documentation won't be wrong for too much longer! ;-P

PR #1405, opened 2021-02-10
update(database config): change well syntax

NathalieC:
* change WARNING and NOTE syntax to check if the wells display better


tbouffard:
This page has been reconverted and the fix is already integrated in the branch.
So closing

PR #1406, opened 2021-02-10
update(hardw softw requ): fix table & well syntax

NathalieC:
* Try to fix the Tables syntax
* Try to fix the well syntax

### DIFF
--- a/modules/ROOT/pages/hardware-and-software-requirements.adoc
+++ b/modules/ROOT/pages/hardware-and-software-requirements.adoc
@@ -12,6 +12,7 @@ processes (number of processes instances, number of current users, operations an
 
 Hardware required for Bonita BPM Platform (Bonita BPM Engine and Bonita BPM Portal)
 
+[width="30%"]
 |===
 | Type | Minimum | Recommended
 
@@ -32,6 +33,7 @@ Hardware required for Bonita BPM Platform (Bonita BPM Engine and Bonita BPM Port
 
 Software required for Bonita BPM Platform (Bonita BPM Engine and Bonita BPM Portal).
 
+[width="30%"]
 |===
 |  | Version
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/hardware-and-software-requirements.adoc
```diff
@@ -12,6 +12,7 @@ processes (number of processes instances, number of current users, operations an
 
 Hardware required for Bonita BPM Platform (Bonita BPM Engine and Bonita BPM Portal)
 
+[width="30%"]
```
tbouffard: `30%` seems too small

**With 30%**
image

**With default**
image

modules/ROOT/pages/hardware-and-software-requirements.adoc
```diff
@@ -32,6 +33,7 @@ Hardware required for Bonita BPM Platform (Bonita BPM Engine and Bonita BPM Port
 
 Software required for Bonita BPM Platform (Bonita BPM Engine and Bonita BPM Portal).
 
+[width="30%"]
```
tbouffard: `30%` seems too small

**With 30%**
image


**With default**
image

tbouffard (changes requested):
I have rebased the PR as the upstream branch had been reconverted
The proposed width for the tables seems too small. As there is no preview for the 7.5 branch, we can work together @NathalieC to find the best table width
We could also set a dedicated width for the first column.

modules/ROOT/pages/hardware-and-software-requirements.adoc
NathalieC: ok thanks. I think that 50% would be nice. I'm trying that.

modules/ROOT/pages/hardware-and-software-requirements.adoc
NathalieC: again, trying 50%

alachambre:
Why do you want to change the table width?? 
Playing with size like that can have a huge impact on mobile view, looks like a bad idea IMHO.

alachambre:
I'm gonna close this pull request, no activity for a month, reopen it if needed.

PR #1414, opened 2021-02-16
add links to videos - 7.10

NathalieC:
* add tv icon


dcoille (approved):
Cf commit with changes

PR #1416, opened 2021-02-17
add links to videos - 7.11

NathalieC:
* add link to BTG videos


benjaminParisel:
I will merge it to avoid the problem for future conversion.

PR #1425, opened 2021-03-01
doc(md): try fix md before conversion to adoc

uguy:
Try fix md before conversion to adoc.

Better markdown means better asciidoc ...


benjaminParisel (approved):
LGTM

PR #1445, opened 2021-03-11
chore(7.11): change the way we trigger the documentation-site workflow

passga:
We are now using a repository_dispatch event to reduce adherence with the
targeted repository.

In addition

* use the new bonita-documentation-site repository name
trigger on content change on all modules changes not only ROOT
covers bonitasoft/bonita-documentation-site#167

* add build pr preview action


tbouffard (approved):
LGTM

PR #1447, opened 2021-03-11
fix(7.9): Fix asciidoc content 

passga:


modules/ROOT/pages/single-sign-on-with-cas.adoc
```diff
@@ -33,84 +33,9 @@ As a consequence, configure the reverse proxy to include the `Host:` header when
 The method https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getRequestURL--[`getRequestURL ot HttpServletRequest`] is used internaly to retrieve the `server` and `port` of the HTTP request.
 ====
 
-==== example
```
tbouffard, line 36: ⚠️ are we sure we want to remove this whole block?

passga: it fix now

PR #1452, opened 2021-03-11
fix(howto): remove uid modal howto

rbioteau:
Deprecated since 7.8 as the UID has a modal widget now


benjaminParisel:
@florentsarat what do you think about removing this how-to. For me, I agree with you Romain, we can remove it.

florentsarat:
@benjaminParisel  @rbioteau Yes we can remove it.

alachambre (approved):

PR #1448, opened 2021-03-11
fix(2021.1): update links to javadoc

tbouffard:
Fix wrongly generated javadoc links because of usage of `$` and `2021.1` (instead of `7.12`) in urls
Also uses the new dedicated domain javadoc.bonitasoft.com and https everywhere.

closes https://github.com/bonitasoft/bonita-documentation-site/issues/191

PR #1456, opened 2021-03-12
fix(7.11): set ref to getting started in index page

tbouffard:
This has already been fixed for 7.9 and 7.10 with #1449 but the getting started home page has changed in 7.11 (see https://github.com/bonitasoft/bonita-doc/pull/1448#discussion_r592548536).


alachambre:
works fine

PR #1457, opened 2021-03-12
fix(7.9): Remove unused icon in Note and Warning admonitions

julienmege:


tbouffard (approved):
LGTM, this is probably not the sole page that have the issue 😄

PR #1458, opened 2021-03-12
Fix declare-business-variables.adoc

florentsarat:
Fixing a few defects after the migration to the nw site

### DIFF
--- a/modules/ROOT/pages/declare-business-variables.adoc
+++ b/modules/ROOT/pages/declare-business-variables.adoc
@@ -37,9 +37,11 @@ Now that a business variable is declared, use it in the transition condition def
 . Click on *<* in the quick access operators' bar and type *3*
 . Click on *OK*
 
-====note
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/declare-business-variables.adoc
```diff
@@ -37,9 +37,11 @@ Now that a business variable is declared, use it in the transition condition def
 . Click on *<* in the quick access operators' bar and type *3*
 . Click on *OK*
 
-====note
+[NOTE]
+====
 You can also enter the Groovy script directly: `claim.satisfactionLevel < 3`
 ====
+[NOTE]
```
tbouffard: this is an extra NOTE, the generated content is ok with it but we should follow the asciidoctor syntax at best to avoid issue in the future 😄 https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/
```suggestion
```

alachambre (approved):

PR #1466, opened 2021-03-12
fix: remove some screenshots

alachambre:
Those screenshots take too much place, and aren't that useful (I don't think people need two screenshot to display a blank page on the web). 

It impact badly the readability of the page, and consume bandwidth for nothing.

Here is a nice article about screenshots in documentation.. http://user-assistance.blogspot.com/2009/10/screen-shots-in-documentation.html


alachambre:
Before removing:

Capture d’écran 2021-03-12 à 17 12 02

PR #1474, opened 2021-03-15
fix(handle-a-failed-activity) fix typo title

marielle-spiteri:
* replace title "Task errors managmenent"  by "Task errors management"
* to be pushed up to latest documentation version


tbouffard (approved):
Good catch, apparently here for a long time 😄

PR #1484, opened 2021-03-22
fix(define-and-deploy-the-bdm.adoc): object name

dcoille:
Fix Appearance issue in the warning about package and object name


alachambre (approved):

alachambre:
You must add a real line break before an admonition, else it doesn't work. I fixed it for you :)

PR #1487, opened 2021-03-22
fix(7.9): fix table in bonita bpm portal urls page

alachambre:
Use asciidoc list instead of non working html tags

### DIFF
--- a/modules/ROOT/pages/bonita-bpm-portal-urls.adoc
+++ b/modules/ROOT/pages/bonita-bpm-portal-urls.adoc
@@ -115,7 +115,7 @@ Example: +
 
 [IMPORTANT]
 ====
-*Important:* A task is not automatically assigned to the user who accesses the task form. There must first be a xref:bpm-api.adoc[REST API call to assign the task to the user].
+A task is not automatically assigned to the user who accesses the task form. There must first be a xref:bpm-api.adoc[REST API call to assign the task to the user].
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/bonita-bpm-portal-urls.adoc
```diff
@@ -199,7 +199,10 @@ With the above format, the first task with the name "request approval" available
 | `locale=en`
 
 | `id=<id>`
-| <ul><li>For process instantiation URL, identifies the process definition id.</li><li>For process overview URL, identifies the process instance id.</li><li>For task URL, identifies the activity instance of the task.</li>
+a|
```
tbouffard, line 202: extra `a`, Typo?
```suggestion
|
```

tbouffard: Answering my own question: **NO** this is not a typo
See https://docs.asciidoctor.org/asciidoc/latest/tables/format-cell-content/#cell-styles-and-their-operators
`a` operator: 
> Supports block elements (lists, delimited blocks, and block macros). This style effectively creates a nested, standalone AsciiDoc document. The parent document’s implicit attributes, such as doctitle, are shadowed and custom attributes are inherited.

tbouffard (commented):
LGTM

PR #1490, opened 2021-03-23
chore(2021.1): introduce a 'bonitaTechnicalVersion' attribute

tbouffard:
This let easily configure the version of the bonita jars used in examples.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -4,6 +4,7 @@ version: 2021.1
 asciidoc:
   attributes:
     varVersion: 2021.1
+    bonitaTechnicalVersion: '7.12.1'
     javadocVersion: 7.12
     page-editable: true
     page-out-of-support: false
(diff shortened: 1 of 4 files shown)


tbouffard:
⚠️ Notice that there are other places where the `varVersion` attribute is currently used and that must be reviewed.
For instance, docker image usage, links to GitHub repositories.

antora.yml
```diff
@@ -4,8 +4,9 @@ version: 2021.1
 asciidoc:
   attributes:
     varVersion: 2021.1
+    bonitaTechnicalVersion: '7.12.1'
```
educhastenier, line 7: Do we need those single quotes?

tbouffard: Adding single quotes forces the value to be considered as a string, otherwise magic happens and it can be considered as a number.
We add issues in the past with version finishing by zero, like `7.10` that were considered as `7.1`. Here, we could remove the quotes because the value currently doesn't end with a zero.
But remember that the change will be propagated to 2021.2; so, we soon will set the value to `7.13.0` in this branch and we may have the same issue (except if we are sure 7.13.0 will be considered as a string).
So I added the single quotes to prevent any problem in the 2021.2 branch.

tbouffard:
@educhastenier we have other area in the 2021.1 documentation where the version must be reviewed (platform/branding or technical) - see https://github.com/bonitasoft/bonita-doc/pull/1490#issuecomment-804765647

PR #1493, opened 2021-03-23
Customer Portal --> Customer Service Center

Guyom-BS:
Tiny detail : we now refer to the Customer Portal as the Customer Service Center


tbouffard:
@Guyom-BS ideally this change should be applied to Bonita 7.9, then it can be propagated to newer Bonita versions like we did in #1467 and #1468. Let's talk about it in person.

Guyom-BS:
To be reopened in older version

PR #1496, opened 2021-03-24
remove(Two Types of Deploy) Completly remove page two-main-types-of-deployment

pabloalonso:
* remove content of unsuported deployment scenario


tbouffard:
@pabloalonso this PR targets bonita 7.5, did you mean targetting bonita 2021.2?
I see a lot of commits to merge into 2021.2 in your branch and the feature you want to remove is deprecated in 2021.1 (and probably in older versions), so I guess removal will happen in 2021.2.

image

preview home also refers to 2021.2

image

Bonita Documentation Preview bonita PR #1496_component_selection

pabloalonso:
@tbouffard 
It is fine to be removed from 7.5. That is the decision has been taken at product level (SoS validated).

abirembaut:
see #1514 instead

PR #1499, opened 2021-03-24
chore(2021.1): fix bonita version attributes substitution

tbouffard:
Fix attributes that were not substituted in code examples (mainly for the Enterprise Docker image).
Introduce dedicated attributes to manage the various versions we have on both components
and tools: this makes things more explicit and provide more flexibility.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -4,10 +4,11 @@ version: 2021.1
 asciidoc:
   attributes:
     bonitaVersion: 2021.1
-    bonitaDockerImageEnterpriseVersion: '7.12.2'
-    crowdinBonitaVersion: 7.12
     bonitaTechnicalVersion: '7.12.1' # latest public community version
     javadocVersion: 7.12
+    bonitaDockerImageEnterpriseVersion: '7.12.2'
+    crowdinBonitaVersion: 7.12
+    # page attributes
     page-editable: true
     page-out-of-support: false
 nav:
--- a/antora.yml
+++ b/antora.yml
@@ -5,8 +5,8 @@ asciidoc:
   attributes:
     varVersion: 2021.1
     bonitaDockerImageEnterpriseVersion: '7.12.2'
-    # latest public community version
-    bonitaTechnicalVersion: '7.12.1'
+    crowdinBonitaVersion: 7.12
+    bonitaTechnicalVersion: '7.12.1' # latest public community version
     javadocVersion: 7.12
     page-editable: true
     page-out-of-support: false
--- a/antora.yml
+++ b/antora.yml
@@ -4,6 +4,8 @@ version: 2021.1
 asciidoc:
   attributes:
     varVersion: 2021.1
+    bonitaDockerImageEnterpriseVersion: '7.12.2'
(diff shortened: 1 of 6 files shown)


tbouffard:
⚠️ Some parts of this Pull Request should also be applied to 7.9+ versions. I have started by 2021.1 as this is the one that requires more changes after the introduction of the branding and technical versions. See also https://github.com/bonitasoft/bonita-doc/pull/1490#issuecomment-804765647
Other branches should only be fixed using the existing `varVersion` asciidoc attribute (the docker usage examples).

In 2021.1, you may also need a new attribute to identify the Entreprise technical version, feel free to replicate the mechanism introduced here.

antora.yml
```diff
@@ -3,9 +3,12 @@ title: Bonita
 version: 2021.1
 asciidoc:
   attributes:
-    varVersion: 2021.1
-    bonitaTechnicalVersion: '7.12.1'
+    bonitaVersion: 2021.1
+    bonitaTechnicalVersion: '7.12.1' # latest public community version
```
educhastenier, line 7:
```suggestion
    bonitaTechnicalVersion: '7.12.1' # latest public community version on this X.Y version line
```

modules/ROOT/pages/bonita-docker-installation.adoc
```diff
@@ -15,18 +15,21 @@ docker run --name bonita -d -p 8080:8080 bonita
 
 To start the latest subscription release
 
-____
-Information on how to access Quay.io can be found https://customer.bonitasoft.com/download/request[here]
 
+// for the 'subs' parameter, see https://docs.asciidoctor.org/asciidoc/latest/subs/apply-subs-to-blocks/
```
educhastenier, line 19: why did you put this information inside the source code, in comments?

educhastenier (approved):
Just a minor remark

antora.yml
tbouffard: I can add this information in the comment
But shouldn't I rename the attribute `bonitaCommunityTechnicalVersion` into to make it more explicit to remove the comment? Like I tried to do for `bonitaDockerImageEnterpriseVersion`

modules/ROOT/pages/bonita-docker-installation.adoc
tbouffard: This is an asciidoctor comment, so it disappears during the html generation.
I added it to let people know about the syntax and explain why it is used. This is not frequent in the documentation, so I thought it's worth mentioning it.
I did it previously in pages where I add this syntax.

If you think this is not relevant, I can remove it, but in that case, I should remove it everywhere in the documentation.

antora.yml
educhastenier: I think not, as I think this version is also used in Enterprise code samples, even if for Enterprise, there may be more recent maintenance versions

modules/ROOT/pages/bonita-docker-installation.adoc
educhastenier: Fine to me

PR #1501, opened 2021-03-26
fix(2021.2): do not mention release dates

tbouffard:
There is no commitment on release dates, so remove them.


pabloalonso (approved):
LGTM
Thanks @tbouffard  for the heads up

PR #1502, opened 2021-03-26
feat(OIDC): support all target URLs

abirembaut:
* add conf parameters to put target URLs in state parameter instead or 
using
redirect_URI when OIDC providers do not support wildcard in valid
redirect URIs
* support default redirect URI when saveTargetURL is true (contextPath + 
/keycloak)
* move resource property to the top as it is one of the most useful properties
* Remove Implicit flow as it is not supported 

Covers 
[RUNTIME-176](https://bonitasoft.atlassian.net/browse/RUNTIME-176)


passga (approved):
LGTM!!!

PR #1513, opened 2021-03-31
feat: add more info on calling apis inside event handler

baptistemesta:


modules/ROOT/pages/event-handlers.adoc
```diff
@@ -45,6 +45,13 @@ What should not be done is:
 * Call Rest APIs: Network call can drastically impact Bonita platform performance.
 * Call Bonita platform Java APIs: It will not work, APIs handle the transaction and when the handler is called, we are already in transaction.
 
+[NOTE]
+====
+
```
alachambre:
```suggestion
```

modules/ROOT/pages/event-handlers.adoc
```diff
@@ -45,6 +45,13 @@ What should not be done is:
...
+If you still want to do Network call of Bonita platform java APIs calls, you can do it by triggering those calls asynchronously using e.g. a message queue or an ExecutorService.
```
alachambre:
```suggestion
If you still want to perform Network call or Bonita platform java APIs calls, you can do it by triggering those calls asynchronously using e.g. a message queue or an ExecutorService. 
```

PR #1520, opened 2021-04-01
fix(SAML) fix formatting

abirembaut:
fix sequence restarting from 1.


pabloalonso:
The section with :
If your Identity Provider is corectly ...
has restarted the numeric list

abirembaut:
@pabloalonso indeed
It shouldn't even be part of the steps.
I remove the number here.
Thanks for the review

PR #1523, opened 2021-04-02
fix: reword Customer Portal --> Customer Service Center

Guyom-BS:
Since January 2021, the Customer Portal should be referred to as the "Customer Service Center"


tbouffard (approved):
Supersedes #1493. Good, as discussed with @Guyom-BS, this should also be done everywhere in the documentation, this one is a good start. The rest will be managed later.
Thanks for contributing.

PR #1534, opened 2021-04-10
Update use-bonita-acm.adoc

smcpoland:
The Note is incorrect and references old editions that no longer exist. Text updates to reflect only Enterprise edition as of 2021.1

### DIFF
--- a/modules/ROOT/pages/use-bonita-acm.adoc
+++ b/modules/ROOT/pages/use-bonita-acm.adoc
@@ -9,7 +9,7 @@ Adaptive Case Management with Bonita offers dynamic modeling of all process poss
 [NOTE]
 ====
 
-For Enterprise, Performance, Efficiency, and Teamwork editions only.
+For Enterprise edition only.
 ====
 
 == Enable ACM Engine Plug-in


tbouffard:
@NathalieC AFAIK we have decided to keep reference to legacy/old edition names in the documentation as some customers are still using plans related to these old editions/subscriptions.
However, the Bonitasoft website don't mention them anymore and this Pull Request shows that this can lead to confusion.

Could we do something about that?
For instance, we could refer to Enterprise only when all Subscriptions/Editions contain a feature and create a page to describe the various Editions (and link to it every time we notice that a feature is Enterprise only); it could mention/explain the old names and that when a feature is flag as Enterprise, it is also included in the legacy subscriptions.
For features only available on some of the old editions and Enterprise, we should find a way to express it.

This will also help to transition the messaging for users/customers, simplify the content and the maintenance (it is a pain to list all subscription types every time).

tbouffard:
@smcpoland thanks for your contribution and helping to improve the Bonita documentation.

For now, we cannot accept it because the change you propose, even if it was considered as valid, should be applied to all pages having references to old editions.
Depending on the @NathalieC answer, we will close this Pull Request or request it to be improved.

smcpoland:
And this is where I humbly disagree on several accounts:

Yes it has to be changed on every page - so what? That's not a good
excuse.  To be honest I would have hoped it would be a templated item that
could be updated once and automatically be shown everywhere.  The only
changes would be to the number of languages the Help is translated into,
i.e. three languages - 3 changes.

2021 and above are exclusively Community and Enterprise - there are no
other versions. Simple.

Help is version specific and therefore changing 2021 to suit it's
specifications does not change any of the other versions text...

Here is the evidence...at 1 we see the directory of the text folder
https://documentation.bonitasoft.com/bonita/2021.1/use-bonita-acm

At 2 we see the Bonitasoft version 2021.1

[image: image.png]

Now change 2 to a different version - any version...in this case 7.10

[image: image.png]

Look the text directory has changed at 1.

And the Text on the page has changed 3.

Go back to 2021.1 and it still looks like 2021.1 above.

The RELEASE NOTES are very clear in this respect, 2021+ is Enterprise and
Community only. Leaving the documentation to reference Teamwork and
Performance is at best lazy - at worst it is wrong.  This is where
customers should learn the difference not elsewhere.  It is wrong and
misleading.


If you must make changes to prior versions of the text - refer to 7.9

This documentation is about a version that is *out of support*, here
is the latest
documentation version <https://documentation.bonitasoft.com/bonita/2021.1/>.

How about downloading <https://www.bonitasoft.com/downloads> a newer,
supported version?



Best regards
Seán

tbouffard:
@smcpoland can you please update your comment https://github.com/bonitasoft/bonita-doc/pull/1534#issuecomment-819203202 directly in the GitHub Web UI? the image you attached to the comment done by sending email are not available.
This will help me to better understand what you have in mind.
Thx

NathalieC:
Hello @smcpoland 
Thanks for highlighting this note.
- You are right saying that we do not sell Teamwork, Efficiency, and Performance editions anymore
- And the features that we add to Bonita are either Community of Enterprise editions, not the old three anymore
- You are right with the analysis of the doc structure: each version comes with its separate set of pages, so we can edit pages on a version with no impact on the same pages of the previous versions
- Indeed, we may have thought about a way to manage all such notes about editions in only one widget to make changes very easy

The fact is that most of our customers still own a Teamwork, Efficiency, or Performance Subscription edition.
As such a customer, reading that a feature is only available for Enterprise edition may mislead me in thinking that I cannot benefit from it.
@tbouffard I agree that we could spend some time writing a page to explain editions, and how the inclusion work between Enterprise and the others. But I have very little hope that customers would go and read it.

So, based on the low ROI that I forsee to create such a page, I'd rather keep the mention of all editions a customer can find the feature in, as long as the majority of our customers still stand in the old edition system.
When the conversion to the new system will be done for most of our customers, we will be happy to lighten the notes about editions to only speak about the one that is valid in the new system: "Enterprise".
I hope it clarifies the standpoint.
Let me know.

smcpoland:
@NathalieC thank you for your thoughts here.

I'm just aware that "Documentation" again is being left to the last when it can be confusing and giving a wrong impression.  I'm currently fighting my own documentation battles with ISO27000 and it's requirements. :)

As a customer I read the release notes, says one thing then read the documentation - says another - I believe every customer will find this confusing and simply revert to the time old tradition of contacting the "Help Desk" to ask what is going on.

Yes, maybe there is a low ROI, but at least customers are not being told two different things.

@tbouffard 's suggestion of a page of edition differences would go a long way to solving this issue, for new, current and past customers (who still watch Bonitasoft).

Anyway, while not happy to give up, at least I've made my point.

Thank you and best regards

NathalieC:
What we can do to avoid the confusion you describe between release notes and documentation pages is to add a sentence in the release notes everytime we talk about an Enterprise feature, to explain that it is also valid if you have any subscription. Will do. 
Thanks @smcpoland

PR #1541, opened 2021-04-12
Fix archive push content action

alachambre:
Update the archive branche push content action, by using the one in the 7.9 branch with the correct parameters


benjaminParisel (approved):
:ok_hand:

PR #1543, opened 2021-04-13
fix(database-conf): detail BDM grants oracle

marielle-spiteri:
* add clear explanation on grants for the BDM DB

### DIFF
--- a/modules/ROOT/pages/database-configuration.adoc
+++ b/modules/ROOT/pages/database-configuration.adoc
@@ -28,7 +28,7 @@ Please note this procedure cannot be performed using the Bonita Studio. The Boni
 To use Bonita on another RDBMS, please use a xref:tomcat-bundle.adoc[bundle] or set up a xref:custom-deployment.adoc[standalone server].
 ====
 
-Here are the steps to follow. They are the same for the engine database and the business data database:
+Here are the steps to follow. They are the same for the Bonita Engine database and the business data database:
 
 . Create the database
 . Customize RDBMS to make it work with Bonita
@@ -59,13 +59,13 @@ Download the driver compatible with Oracle 12.2.0.1: https://www.oracle.com/data
 
 === Create the database(s)
 
-The first step in configuring Bonita with your RDBMS is to create the new databases (i.e. schemas): one for engine data, and another one, optionally, if you use business data.
+The first step in configuring Bonita with your RDBMS is to create the new databases (i.e. schemas): one for the Bonita Engine data, and another one, optionally, if you use business data.
 
-WARNING: If you use different users for the engine database and for the business data database, then you must *grant the following privileges to both users*. 
+WARNING: If you use different RDBMS users for the Bonita Engine database and for the business data database, then you must *grant the privileges to both users*. 
 
-To create each new database, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schema).
+To create each Bonita Engine and business data database, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create each new schemas).
 
-Next, the owner of each new schema (i.e. engine data *and* business data) must own the following privileges:
+Next, the RDBMS user owner of each new schema (i.e. Bonita Engine data *and* business data) must own the following privileges:
 
 * CREATE TABLE
 * CREATE INDEX
@@ -74,8 +74,8 @@ Next, the owner of each new schema (i.e. engine data *and* business data) must o
 [NOTE]
 ====
 
-CREATE TABLE and CREATE INDEX privileges are not required for the *engine database* user after first start in normal use (however, they will still be required by the *business data database* user even beyond the first deployment of a BDM). +
-If the same SQL user is used with the xref:migrate-from-an-earlier-version-of-bonita-bpm.adoc[migration tool], then this user needs such grants.
+CREATE TABLE and CREATE INDEX privileges are not required for the *Bonita Engine database* RDBMS user after first start in normal use (however, they will still be required by the *business data database* user even beyond the first deployment of a BDM). +
+If the same RDBMS user is used with the xref:migrate-from-an-earlier-version-of-bonita-bpm.adoc[migration tool], then this user needs such grants.
 ====
 
 To create the database(s), we recommend that you refer to your RDBMS documentation:
@@ -91,8 +91,8 @@ Also, you are recommended to configure the database(s) to be case-insensitive so
 
 [#specific_database_configuration]
 
-NOTE: Bonita is not attached to a specific database user. Therefore, the database user that bonita uses to connect to the database can be changed after the platform creation, provided the new user has the required access rights.
-The same goes for the user used to connect to the business data database, if it is different.
+NOTE: Bonita is not attached to a specific RDBMS user. Therefore, the RDBMS user that bonita uses to connect to the database can be changed after the platform creation, provided the new user has the required access rights.
+The same goes for the RDBMS user used to connect to the business data database, if it is different.
 
 === Customize RDBMS to make it work with Bonita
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/database-configuration.adoc
```diff
@@ -61,9 +61,11 @@ Download the driver compatible with Oracle 12.2.0.1: https://www.oracle.com/data
 
 The first step in configuring Bonita with your RDBMS is to create the new databases (i.e. schemas): one for engine data, and another one, optionally, if you use business data.
 
-To do so, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schema).
+WARNING: If you use different users for the engine database and for the business data database, then you must *grant the following privileges to both users*.
```
NathalieC: This warning should come below the first sentence  that mentions the users.

modules/ROOT/pages/database-configuration.adoc
```diff
@@ -61,9 +61,11 @@ Download the driver compatible with Oracle 12.2.0.1: https://www.oracle.com/data
 
 The first step in configuring Bonita with your RDBMS is to create the new databases (i.e. schemas): one for engine data, and another one, optionally, if you use business data.
 
-To do so, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schema).
+WARNING: If you use different users for the engine database and for the business data database, then you must *grant the following privileges to both users*. 
 
-Also, note that the owner of the new schemas must own the following privileges:
+To create each new database, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schema).
```
NathalieC:
```suggestion
To create each new database, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schemas).
```

modules/ROOT/pages/database-configuration.adoc
NathalieC:
```suggestion
WARNING: If you use different RDBMS users for the Bonita Engine database and for the business data database, then you must *grant the following privileges to both users*. 
```

NathalieC: This sentence talks about "the following privileges". 
Row 68 also talks about "the following privileges".
But this one is about the RDBMS users, and the other one is about a database owner. 
- Are those users the same? If so, I'd be in favor of always referring to them with the same descriptor.
- Can we talk only once about "the following privileges" (the one time that is closer to the description of the privileges)?

modules/ROOT/pages/database-configuration.adoc
```diff
@@ -187,7 +189,12 @@ SQL query >
 @/u01/app/oracle/product/12.2.0/dbhome_1/rdbms/admin/xaview.sql
 ----
 +
-The following queries must be done for each new user: i.e. one user = one database schema.
+The following queries must be done for each new user: i.e. one user = one database schema.  
++
+WARNING: The queries must be done for *both* the bonita engine *and* the business data database users. +
+For example, if the users for the Bonita BPM and BDM databases are `bonita` and `business_data` respectively (the actual values correspond to the ones set in the properties `db.user` and the `bdm.db.user` defined in the `<TOMCAT_HOME>/setup/database.properties`), you need to run the queries for both users.
```
NathalieC:
```suggestion
For example, if the users for the Bonita Engine and business data databases are `bonita` and `business_data` respectively (the actual values correspond to the ones set in the properties `db.user` and the `bdm.db.user` defined in the `<TOMCAT_HOME>/setup/database.properties`), you need to run the queries for both users. 
```

modules/ROOT/pages/database-configuration.adoc
```diff
@@ -187,7 +189,12 @@ SQL query >
 @/u01/app/oracle/product/12.2.0/dbhome_1/rdbms/admin/xaview.sql
 ----
 +
-The following queries must be done for each new user: i.e. one user = one database schema.
+The following queries must be done for each new user: i.e. one user = one database schema.  
++
+WARNING: The queries must be done for *both* the bonita engine *and* the business data database users. +
+For example, if the users for the Bonita BPM and BDM databases are `bonita` and `business_data` respectively (the actual values correspond to the ones set in the properties `db.user` and the `bdm.db.user` defined in the `<TOMCAT_HOME>/setup/database.properties`), you need to run the queries for both users. 
++
+*For the bonita engine database user (e.g. _bonita_):*
```
NathalieC:
```suggestion
*For the Bonita Engine database user (e.g. _bonita_):*
```

modules/ROOT/pages/database-configuration.adoc
```diff
@@ -187,7 +189,12 @@ SQL query >
 @/u01/app/oracle/product/12.2.0/dbhome_1/rdbms/admin/xaview.sql
 ----
 +
-The following queries must be done for each new user: i.e. one user = one database schema.
+The following queries must be done for each new user: i.e. one user = one database schema.  
++
+WARNING: The queries must be done for *both* the bonita engine *and* the business data database users. +
```
NathalieC:
```suggestion
WARNING: The queries must be done for *both* the Bonita Engine *and* the business data database users. +
```

modules/ROOT/pages/database-configuration.adoc
```diff
@@ -72,7 +74,7 @@ Also, note that the owner of the new schemas must own the following privileges:
 [NOTE]
 ====
 
-CREATE TABLE and CREATE INDEX privileges are not required after first start in normal use. +
+CREATE TABLE and CREATE INDEX privileges are not required for the *engine database* user after first start in normal use (however, they will still be required by the *business data database* user even beyond the first deployment of a BDM). +
```
NathalieC: We cannot say that the owners of both databases must have the same privileges and then say that some privileges are not required for one of them.
For more clarity, I would suggest to first describe the privileges of the Bonita Engine database owner, and then the privileges of the business data database.

modules/ROOT/pages/database-configuration.adoc
```diff
@@ -61,9 +61,11 @@ Download the driver compatible with Oracle 12.2.0.1: https://www.oracle.com/data
 
 The first step in configuring Bonita with your RDBMS is to create the new databases (i.e. schemas): one for engine data, and another one, optionally, if you use business data.
 
-To do so, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schema).
+WARNING: If you use different users for the engine database and for the business data database, then you must *grant the following privileges to both users*. 
 
-Also, note that the owner of the new schemas must own the following privileges:
+To create each new database, you need a RDBMS user account that has sufficient privileges (i.e. privileges to create new schema).
+
+Next, the owner of each new schema (i.e. engine data *and* business data) must own the following privileges:
```
NathalieC:
```suggestion
Next, the owner of each new schema (i.e. Bonita Engine data *and* business data) must own the following privileges:
```

modules/ROOT/pages/database-configuration.adoc
NathalieC:
```suggestion
CREATE TABLE and CREATE INDEX privileges are not required for the *Bonita Engine database* user after first start in normal use (however, they will still be required by the *business data database* user even beyond the first deployment of a BDM). +
```

NathalieC (changes requested):

modules/ROOT/pages/database-configuration.adoc
educhastenier: And instead of **Engine**, I would say **Runtime**

educhastenier (approved):
Just one global remark: in this page, the naming of the "Bonita" / "Engine" database is not always named the same way.
Please homogeneize it (have a look at its name on other pages as well)

modules/ROOT/pages/database-configuration.adoc
marielle-spiteri: Change done

marielle-spiteri: * I tried to hamronize the user descriptor throughout the page where it made sense. 
* removed following from the first sentence.

modules/ROOT/pages/database-configuration.adoc
marielle-spiteri: * I agree with @NathalieC , this note is not useful and there are too many conditions for these priviledges aren't required. But 90% of the time, they are required (except when "first start in normal use" which is not clear to me, please define "first start in normal use"...). Let's say they are always required so we cover all the use cases. And we can remove that note. 
* Regarding the Engine vs Runtime, is that the new word to be used in the documentation? If that is the case, then we need to review all the pages that talk about the Engine. I say Engine is fine for me.

educhastenier: @mariellebonita Fine to me.
I let you do this last removal and merge the PR like that (the rest is fine to me).

educhastenier: @mariellebonita I removed the note

PR #1509, opened 2021-03-30
update(purge-tool): Add warn document not deleted

marielle-spiteri:
* Document not deleted warning

PR #1545, opened 2021-04-14
Update(purge-tool): warning documents not deleted

marielle-spiteri:
* Adding a clear message that the Documents aren't deleted from the database
* If the user needs to reduce the size of the document table, they need to use the Java APIs


marielle-spiteri:
I have closed the #1509, @pabloalonso could you please validate this warning for @alachambre  , so that this big limitation (RUNTIME-147) is documented for our subscription users to know what to expect. Thanks in advance

pabloalonso (approved):
LGTM

PR #1553, opened 2021-04-19
add(bdm): warning use lazy not eager

marielle-spiteri:
* Use lazy not eager warning

### DIFF
--- a/modules/ROOT/pages/define-and-deploy-the-bdm.adoc
+++ b/modules/ROOT/pages/define-and-deploy-the-bdm.adoc
@@ -152,11 +152,6 @@ When you create a process that uses a business object with a composition or aggr
 [discrete]
 ==== Loading
 
-[WARNING]
-====
-Bonita strongly recommends to use the *lazy* loading option. The eager loading option has a huge impact on the platform performances. You should only use it if you really need all the child object instances to be loaded when the parent object instance is loaded. If you decide to use the eager mode, we recommend to test this option with a test dataset that is the size of your production. 
-====
-
 There are two options for loading complex business objects, *lazy* or *eager*.
 With lazy loading, the parent object instance is loaded, but child object instances are loaded only when they are needed.
 With eager loading, child object instances are loaded when the parent object instance is loaded.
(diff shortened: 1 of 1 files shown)


NathalieC:
Hello,
I think the warning should come after the definition of both modes.
Thanks for adding it.

marielle-spiteri (commented):
I moved the warning after description

NathalieC:
I just saw that before the warning you have added, another warning already exists, less visible. Those two need to be merged.

NathalieC:
I merged the two warnings into one and made some editions after speaking with Poorav, NicoChi, and Florent.
If you agree, I'll merge the PR.

modules/ROOT/pages/define-and-deploy-the-bdm.adoc
```diff
@@ -152,14 +152,19 @@ When you create a process that uses a business object with a composition or aggr
...
-
-The default is lazy loading. You can override this for any business object by configuring it to use eager loading when you specify
-the object relationship properties. This means that the specified object and all objects related to it by composition or aggregation
-relationships are loaded. +
-There is an overhead for the additional data load and maintenance of the information in memory, but there is a saving in data access time because the data is already loaded.
+There are two options for loading complex business objects, *lazy* or *eager*. +
+
+* With lazy loading, the parent object instance is loaded, and child object instances are loaded only when they are needed, by the means of additional REST API calls. This is the option to choose when the parent object along with all its relationships can get big, vertically and/or horizontally +
+* With eager loading, the child object instances are loaded when the parent object instance is loaded. It is easier to use because no additional REST API calls are needed. This is the option to choose when the parent object along with all its relationships are neither deep nor large.  +
```
florentsarat: The lazy or eager choice should made based on the use case. 

Objects that are rarely changed or displayed > Lazy
Objects that are changed or display often > Eager

Even if the object is big, if it has be changed often - it will require more coding lines for the dev and in the end the performance impact will be the same as you will need to load everything that you change

marielle-spiteri: @florentsarat, can you please clarify your comment, I don't understand it. Do you want to change this documentation update?

florentsarat: Yes, we need to change it. The lazy/eager choice shouldn't be decided solely based on the object size. How the object is going to be used should also be taken into account.

marielle-spiteri: Ok no problem to rephrase as long as this real warning having a critical impact on many customers gets published eventually ;-) 
I let you do the rephrase based on your offline discussions with @NathalieC, @pooravbos, Nicolas Chignardet and other. Thanks a lot!

NathalieC: I'm adding Florent's comment and will merge the PR.

PR #1554, opened 2021-04-20
feat(call): info on called process with no version specified

baptistemesta:
When no version is specified, the call activity takes the last version deployed of that process

### DIFF
--- a/modules/ROOT/pages/called-processes.adoc
+++ b/modules/ROOT/pages/called-processes.adoc
@@ -33,7 +33,7 @@ Define the called process as you would define any other process. When you create
 In the parent process, define a call activity task. Define the normal task characteristics. Note that although a call activity cannot
 have forms associated with it, it can have connectors and task-level data.
 
-In the *Details* panel, *General* tab, *General* pane, specify the name of the called process. If you know the process name, you can enter it directly, otherwise use the expression editor.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/called-processes.adoc
```diff
@@ -33,7 +33,7 @@ Define the called process as you would define any other process. When you create
 In the parent process, define a call activity task. Define the normal task characteristics. Note that although a call activity cannot
 have forms associated with it, it can have connectors and task-level data.
 
-In the *Details* panel, *General* tab, *General* pane, specify the name of the called process. If you know the process name, you can enter it directly, otherwise use the expression editor.
+In the *Details* panel, *General* tab, *General* pane, specify the name of the called process. If you know the process name, you can enter it directly, otherwise use the expression editor. The version is optional, when not specified, the engine takes the last version deployed of the process.
```
ugaston:
```suggestion
In the *Details* panel, *General* tab, *General* pane, specify the name of the called process. If you know the process name, you can enter it directly, otherwise use the expression editor. 

[NOTE]
====

The version is optional, when not specified, the engine takes the last version deployed of the process.
====
```

ugaston (commented):
Looks fine.
But what about putting the information in a NOTE, so it doesn't get lost in the forest of monotonous lines? ;)

PR #1556, opened 2021-04-28
Add project best practices, remove Project creation, edit Quick Start and runtime install

NathalieC:
* add project best practices top level chapter
* add dedicated index page
* reference 2 existing pages
* remove ProjectCreationIndex page
* move its content to the top level


modules/ROOT/pages/project-best-practicesIndex.adoc
```diff
@@ -0,0 +1,26 @@
+= Project best practices
```
alachambre, line 1: I would prefer you to use camel case for the file name instead of kebab case, it's better to stick with the convention unless you are able to justify this choice :)

FYI: https://wprock.fr/blog/conventions-nommage-programmation/

modules/ROOT/pages/project-best-practicesIndex.adoc
```diff
@@ -0,0 +1,26 @@
...
+--
+
+[.card.card-index]
+--
+xref:design-methodology.adoc[[.card-title]#Conception# [.card-body.card-content-overflow]#pass:q[TODO]#]
+--
+
+[.card.card-index]
+--
+xref:setupDevEnvironmentIndex.adoc[[.card-title]#Set up the development environment# [.card-body.card-content-overflow]#pass:q[]#]
```
alachambre, line 20: I don't think that the set up documentation is really related with project best practices. By that I mean that I would never look into the best practices if i was looking for a guide to setup my environment. I would probably expect it to be in the quick start.

NathalieC: ok thx

modules/ROOT/pages/project-best-practicesIndex.adoc
NathalieC: oops. Thx

PR #1564, opened 2021-05-05
fix(release notes): change release date

nchabanoles:
The release happened in 2021 not in 2020


NathalieC:
Good you checked the release dates of the maintenance versions ;)
Thanks for this catch.

PR #1571, opened 2021-05-11
update(taxo and old taxo) : howto and quick start updates

NathalieC:
* move howtos in doc


tbouffard:
⚠️ @NathalieC be careful when renaming files existing in oldest versions (if the file has been created in the current version, this is not an issue) like you do in this Pull Request for 
- BonitaBPM_platform_setup: https://documentation.bonitasoft.com/bonita/2021.1/bonitabpm_platform_setup
- profileCreation: https://documentation.bonitasoft.com/bonita/2021.1/profilecreation

When you do this, don't forget to declare an alias to renamed page as this as an impact on the SEO. https://github.com/bonitasoft/bonita-documentation-site/blob/a080cea1f8323ae9442c4397d9eed0d59bfb4419/docs/content/CONTRIBUTING.adoc#alias-when-renaming-pages
This should also allow to link the page to other version (to be confirmed): the version drop down list may allow to navigate between old and new version even if the page has been renamed.

If you already did it in other Pull Request, the impacted pages must be updated as well.

If this Pull Request is only a playground, my comment will be at least a reminder 😄

PR #1574, opened 2021-05-17
feat(BDM): document the runtime implementation of BDM deployment

educhastenier:
Notably, the BDM class generation and the BDM database creation/update


educhastenier (commented):
@danila-m Shouldn't we indicate the manual operations to work around the limitations?

PR #1587, opened 2021-06-10
fix(links): fix links that were valid in the previous doc

educhastenier:
before the new Doc infra & tooling refactoring

### DIFF
--- a/modules/ROOT/pages/single-sign-on-with-kerberos.adoc
+++ b/modules/ROOT/pages/single-sign-on-with-kerberos.adoc
@@ -209,11 +209,11 @@ edit the authenticationManager-config.properties as follows:
 ----
 +
 
-Make sure to link:multi-tenancy-and-tenant-configuration#_tenant_creation[set the right tenant admin username].
+Make sure to link:multi-tenancy-and-tenant-configuration[set the right tenant admin username].
 It is recommended to also replace the value of the passphrase (property auth.passphrase) which is used by the engine to verify the authentication request.
 The value must be the same as in the file *bonita-tenant-sp-custom.properties*. +
 If the users need to bypass kerberos authentication method, you can authorize it by setting the property `kerberos.auth.standard.allowed` to true. Users will then be able to log in using the portal login page (/login.jsp) provided they have a bonita account and their password is different from their username. +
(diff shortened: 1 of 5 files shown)


modules/ROOT/pages/single-sign-on-with-kerberos.adoc
```diff
@@ -208,7 +208,7 @@ edit the authenticationManager-config.properties as follows:
      -->	logout.link.hidden=true
 ----
 +
-Make sure to link:multi-tenancy-and-tenant-configuration#toc2[set the right tenant admin username].
+Make sure to link:multi-tenancy-and-tenant-configuration[set the right tenant admin username].
```
educhastenier, line 212: This paragraph does not exist anymore, as the multi-tenancy is now deprecated. I am not sure of what to do with this link

modules/ROOT/pages/single-sign-on-with-saml.adoc
```diff
@@ -93,7 +93,7 @@ edit the authenticationManager-config.properties as follows:
         logout.link.hidden=true
 ----
 +
-Make sure to link:multi-tenancy-and-tenant-configuration#toc2[set the right tenant admin username].
+Make sure to link:multi-tenancy-and-tenant-configuration[set the right tenant admin username].
```
educhastenier: This paragraph does not exist anymore, as the multi-tenancy is now deprecated. I am not sure of what to do with this link

tbouffard:
ℹ️ This link containing `toc` anchor are probably wrong in 7.9+ branches as well.

educhastenier:
>  This link containing `toc` anchor are probably wrong in 7.9+ branches as well.

Yes, for some of them, but from one branch to another, some more were added, so to speed up the fix, I did it on the 2021.1 branch only.
Another reason is that I found this bug when working on another subject, so I did not want to dedicate too much time to it.

However, if you think this is important to fix it for all 7.9+, I can take time to do it properly.

educhastenier:
Ok, I did it on 7.9 as well: https://github.com/bonitasoft/bonita-doc/pull/1589

Not sure I will do it also on 7.10, 7.11, though

PR #1595, opened 2021-06-15
update(ldap sync): add search parameters info

marielle-spiteri:
* add info about Paged search parmeters
use_paged_search = false
page_size = 1000


ugaston (approved):
Brilliant!
Now customers will now ahead about the existence of those parameters.
Thanks!!

PR #1601, opened 2021-06-21
fix(bdm_limitations): change type attribute not supported

marielle-spiteri:
* add Changing the type of an attribute not supported


educhastenier:
Looks good to me :+1:

PR #1602, opened 2021-06-23
rename(bonita studio pages): use kebab case

NathalieC:
* rename page to use kebab base


alachambre:
Did you check in the whole project (using the Github search for ex, not manually!) that those pages were not referenced somewhere else?

florentsarat:
Filenaming was fixed in another PR after your comments in the Product weekly- closing the PR

PR #1614, opened 2021-07-06
chore: remove outdated installation page

baptistemesta:
Also move "Install Tomcat with Bonita as a service in Windows"
at the end of the "Advanced installation section"

The page "Install Tomcat with Bonita as a service in Windows" will also
be updated to remove outdated informations and point the the existing
installation procedure

Relates to [BS-19468](https://bonitasoft.atlassian.net/browse/BS-19468)


pabloalonso (approved):
Lgtm

PR #1646, opened 2021-07-29
feat(com/ent upgrade): Change where to check the upgrade

DumitruCorini:
Remove mention of portal since there will be no portal in 2021.2


educhastenier (approved):
Looks good to me :+1:

alachambre (approved):

PR #1665, opened 2021-08-19
Move process testing overview content to the index

NathalieC:
* add the content of the deleted overview

### DIFF
--- a/modules/ROOT/pages/process-testing-index.adoc
+++ b/modules/ROOT/pages/process-testing-index.adoc
@@ -1,5 +1,5 @@
 = Process testing 
-:description: 
+:description: Tips and how-tos for functional and performance testing of your automation projects.
 
 {description} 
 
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/process-testing-index.adoc
```diff
@@ -1,31 +1,46 @@
 = Process testing 
-:description: blabla
+:description:
```
alachambre: You must add a short description with keywords for SEO. If you don't want this description to be visible then dont add the {description} after, but ensure that keywords will be catch by the SEO robots.

alachambre (dismissed):

modules/ROOT/pages/process-testing-index.adoc
NathalieC: Good catch. Thx.
Done.

PR #1679, opened 2021-08-26
fix(ldap-sync) fix format section troubleshooting

marielle-spiteri:
* add troubleshooting icon
* remove anchors and format sections with the same paterns as others troubleshooting section in the other pages
* Needs to be pushed up to latest doc version


educhastenier:
> * Needs to be pushed up to latest doc version

Hi @mariellebonita.
There is no need to precise this.
It is always the case.
All you have to do is to base your PR on the first version where the changes need to be done, and our **internal tooling** will take care of the rest 😉

educhastenier (approved):
Looks good to me :+1:

PR #1691, opened 2021-08-31
doc(BDM in cluster): specify the BDM in cluster

educhastenier:
More precisely, specify that there is nothing to do
to make the BDM update work in a cluster.


baptistemesta:
Does this come from customer/community feedback?
I'm not sure it is that useful to add something saying that there is nothing special regarding cluster deployment. 
IMO Cluster deployment should not behave differently than a single node deployment by default and we should only specify when there is differences

educhastenier:
> Does this come from customer/community feedback?
> I'm not sure it is that useful to add something saying that there is nothing special regarding cluster deployment.
> IMO Cluster deployment should not behave differently than a single node deployment by default and we should only specify when there is differences

@baptistemesta It's coming from a "Community" question: https://community.bonitasoft.com/questions-and-answers/how-should-i-proceed-install-bdm-bonita-cluster

PR #1702, opened 2021-09-13
Update taxonomy.adoc

NathalieC:
* add link to configuration index


NathalieC:
will make this change in another PR, in order to reduce the number of PR.

PR #1703, opened 2021-09-13
Update taxonomy.adoc

NathalieC:
* move some overviews at the level of indexes


modules/ROOT/taxonomy.adoc
```diff
@@ -181,21 +181,19 @@
...
-  *** Profiles
-   **** xref:profiles-overview.adoc[Overview]
+  *** xref:profiles-overview.adoc[Profiles]
    **** xref:profile-creation.adoc[Profile editor]
   *** xref:rest-api-authorization.adoc[REST API authorization]
   *** xref:custom-authorization-rule-mapping.adoc[Authorization rules]
   *** xref:bdm-access-control.adoc[BDM access control]
- ** Authentication
-  *** xref:user-authentication-overview.adoc[User authentication overview]
+ ** [Authentication]
```
alachambre: I don't understand this change, and I disagree: it looks weird and doesn't make sens for users.

```suggestion
 ** Authentication
```

NathalieC: I do not understand what you disagree with ; can you be more specific, and explain what you would rather have?

PR #1742, opened 2021-09-28
fix(workspace): add troubleshooting for missing projects after crash

rbioteau:
Closes STUDIO-4118

### DIFF
--- a/modules/ROOT/pages/workspaces-and-repositories.adoc
+++ b/modules/ROOT/pages/workspaces-and-repositories.adoc
@@ -298,3 +298,10 @@ To restore an older point:
 . Go to the *Team* > *SVN*, then *Manage restore points*.
 . Select *Restore to an older point* then click *_Next_*.
 . Select the restore point you want to restore, then click *_Restore_*. The project reverts to the restore point.
+
+== image:images/troubleshooting.png[troubleshooting-icon] Troubleshooting
+
+NOTE: Since 2021.2-u1
+
+**My Studio crashed and I've lost my projects in my workspace.** +
+You can enabled the workspace recovery mode that will synchronize the file system and the Studio workspace. Open the `BonitaStudio*.ini` file matching the executable name and set the value of the system property `workspace.recovery.mode` to `true`. Then, restart the Studio. In the Studio logs you should see that the recovery mode is enabled and the import operation of the missing projects into the workspace. Once the recovery operation done, you may want to disable the workspace recovery mode by resetting the property value to `false` in the `.ini` file.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/workspaces-and-repositories.adoc
```diff
@@ -298,3 +298,10 @@ To restore an older point:
 . Go to the *Team* > *SVN*, then *Manage restore points*.
 . Select *Restore to an older point* then click *_Next_*.
 . Select the restore point you want to restore, then click *_Restore_*. The project reverts to the restore point.
+
+== image:images/troubleshooting.png[troubleshooting-icon] Troubleshooting
+
+NOTE: Since 2021.2-u1
```
alachambre, line 304:
```suggestion
```
Useless, you add this on the 2021.1 branch, older versions won't have the section.

modules/ROOT/pages/workspaces-and-repositories.adoc
```diff
@@ -298,3 +298,10 @@ To restore an older point:
...
+
+**My Studio crashed and I've lost my projects in my workspace.** +
```
alachambre: I suggest to use a lvl 3 title here instead of a blod title -> a real title appears in the table of content and shows up quickly with the search. It's what you want for troublshooting
```suggestion
=== My Studio crashed and I've lost my projects in my workspace
```

modules/ROOT/pages/workspaces-and-repositories.adoc
```diff
@@ -298,3 +298,10 @@ To restore an older point:
...
+You can enabled the workspace recovery mode that will synchronize the file system and the Studio workspace. Open the `BonitaStudio*.ini` file matching the executable name and set the value of the system property `workspace.recovery.mode` to `true`. Then, restart the Studio. In the Studio logs you should see that the recovery mode is enabled and the import operation of the missing projects into the workspace. Once the recovery operation done, you may want to disable the workspace recovery mode by resetting the property value to `false` in the `.ini` file.
```
alachambre:
```suggestion
You can enabled the workspace recovery mode that will synchronize the file system and the Studio workspace. +
Open the `BonitaStudio*.ini` file matching the executable name (in the bonita installation folder, next to the workspace folder) and set the value of the system property `workspace.recovery.mode` to `true`. Then, restart the Studio. +
In the Studio logs, you should see that the recovery mode is enabled and the import operation of the missing projects into the workspace. Once the recovery operation done, you may want to disable the workspace recovery mode by resetting the property value to `false` in the `.ini` file.
```

alachambre (changes requested):

modules/ROOT/pages/workspaces-and-repositories.adoc
rbioteau, line 307: It won't be available in 2021.2-u0, so I need to explicitly specify the maintenance version, no ?

alachambre (approved):

PR #1747, opened 2021-09-30
chore: skip preview for PR created from fork repositories

tbouffard:
This prevents deployment errors: there is no more deployment but the PR creator won't think that something is broken in the documentation.

covers https://github.com/bonitasoft/bonita-documentation-site/issues/239

### Tests done

2328a71 simulates unavailable secret
4b9dd55 shows that the preview is built when the secret is available

### DIFF
--- a/.github/workflows/build-pr-preview.yml
+++ b/.github/workflows/build-pr-preview.yml
@@ -5,7 +5,7 @@ on:
     # To manage 'surge-preview' action teardown, add default event types + closed event type
     types: [opened, synchronize, reopened, closed]
     paths:
-      - 'modules/ROOT/**'
+      - 'modules/**'
       - 'antora.yml'
       - '.github/workflows/build-pr-preview.yml'
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-pr-preview.yml
```diff
@@ -5,11 +5,13 @@ on:
     # To manage 'surge-preview' action teardown, add default event types + closed event type
     types: [opened, synchronize, reopened, closed]
     paths:
-      - 'modules/ROOT/**'
+      - 'modules/**'
```
tbouffard, line 8: prepare usage of several Antora modules

.github/workflows/build-pr-preview.yml
```diff
@@ -5,11 +5,13 @@ on:
...
       - 'antora.yml'
       - '.github/workflows/build-pr-preview.yml'
 jobs:
   build_preview:
+    # Don't build preview for non member of the repository
+    if: ${{ secrets.SURGE_TOKEN_DOC != '' }}
```
tbouffard: This doesn't work because here, the workflow cannot access to the env or secrets context
We will run another job that will expose a boolean as job output, and the output will be available here

alachambre (approved):

PR #1750, opened 2021-10-04
Delete connectors-overview.adoc

florentsarat:


tbouffard (changes requested):
Please put an alias on the page that replaces this one.
More details in https://github.com/bonitasoft/bonita-documentation-site/blob/master/docs/content/CONTRIBUTING.adoc#alias-when-renaming-pages

tbouffard (approved):
Thanks for the alias addition 😸

PR #1751, opened 2021-10-05
fix(RN): Reorder the releases in the release note

danila-m:


pabloalonso (commented):
is it the right possition?

It should be between...
"Bug fixes" and "Fixes in Bonita 2021.1-0811 (2021-08-11)"

modules/ROOT/pages/release-notes.adoc
```diff
@@ -233,7 +233,43 @@ In the Community edition, XPDL and jBPM importers have been removed. +
...
+* Jackson: `2.11.2` -> `2.12.3`
+* Javassist: `3.18.1-GA` -> `3.27.0-GA`
+* XStream: `1.4.15` -> `1.4.17`
+* Ehcache: `2.10.6` -> `2.10.9.2`
+* javax.servlet-api: `3.0.1` -> `3.1.0`
+* jakarta.servlet.jsp.jstl 1.2.5 -> 1.2.6
+* jakarta.servlet.jsp.jstl-api 1.2 -> 1.2.7
+* org.apache.taglibs: `1.1.2` -> `1.2.5`
+* com.thoughtworks.xstream: `1.4.17` -> `1.4.18`
+* postgresql driver: `42.2.5` -> `42.2.13`
```
tbouffard: I suggest you precise in which version the updates have been done.
Reading this without other information, I have the feeling this applies to all versions and changes are then harder to track.

I guess the idea of moving the library updates here is that it was previously in the 'bug fixes' paragraph and you may not consider lib updates as bug fixes.
An alternative could be to not only have a bug fixes paragraph but something related to the maintenance (or any name you want).  The maintenance paragraph then can contains subparagraphs: bug fixes, lib update, other changes, ....

danila-m: I agree on principle.
Thing is, there was no separation by maintenance releases before this PR (I regrouped 2 sections, but there were not exhaustive in any way)...

I can do what you suggest by going through the engine/web git tags if more people think it should be done.

alachambre:
The release note file hasn't moved during the switch to modules so you do not have to recreate this pull request, but I suggest you to merge it quickly or at least to rebase it, to avoid mistakes.

modules/ROOT/pages/release-notes.adoc
pabloalonso: I agree with @tbouffard 
Before the lib updates where below the title of the maintenance release in which it was modified.

we can do as example 
==== Library updates
===== since 2021.1-xxyy
* Spring framework: `5.2.9.RELEASE` -> `5.2.15.RELEASE`
* Spring boot: `2.3.4.RELEASE` -> `2.3.12.RELEASE`
* Jackson: `2.11.2` -> `2.12.3`
* Javassist: `3.18.1-GA` -> `3.27.0-GA`
* XStream: `1.4.15` -> `1.4.17`
* Ehcache: `2.10.6` -> `2.10.9.2`
* javax.servlet-api: `3.0.1` -> `3.1.0`
* org.apache.taglibs: `1.1.2` -> `1.2.5`
* com.thoughtworks.xstream: `1.4.17` -> `1.4.18`
* postgresql driver: `42.2.5` -> `42.2.13`
===== since 2021.1-aabb
* jakarta.servlet.jsp.jstl 1.2.5 -> 1.2.6
* jakarta.servlet.jsp.jstl-api 1.2 -> 1.2.7

pabloalonso (approved):
LGTM

PR #1790, opened 2021-10-15
feat(runtime doc): Create engine-deep-dive index page

alachambre:
- Update navigation accordingly
- Add missing line in the runtime navigation (got manually removed during migration to module because it references a page in an other module, I forgot to add it again)


modules/runtime/pages/engine-deep-dive-index.adoc
```diff
@@ -0,0 +1,67 @@
...
+:description: This page describes the core elements to know about in Bonita Engine to start mastering what goes on under Bonita's hood.
+
+{description}
+
+[.card-section]
+== Bonita under the hood
+
+[.card.card-index]
+--
+xref:engine-architecture-overview.adoc[[.card-title]#Architecture overview# [.card-body.card-content-overflow]#pass:q[]#]
```
alachambre, line 11: careful this page is in an other module now -> `xref:module:page`

NathalieC: it should be in this one. The first reference to it should be modified. 
I'll move it and change the reference in the "components" page.

NathalieC: Thank you @alachambre

PR #1794, opened 2021-10-19
fix(update Runtime): database table

NathalieC:
* fix syntax for database table

### DIFF
--- a/modules/version-update/pages/migrate-from-an-earlier-version-of-bonita.adoc
+++ b/modules/version-update/pages/migrate-from-an-earlier-version-of-bonita.adoc
@@ -155,8 +155,9 @@ The correct driver is mandatory for a smooth update: xref:ROOT:database-configur
 . Configure the database properties needed by the migration script, by editing `bonita-update/Config.properties`. +
 Specify the following information:
 
+  [cols="1,1,1"]
   |===
-  | Property | Description | Example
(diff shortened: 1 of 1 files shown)


modules/version-update/pages/migrate-from-an-earlier-version-of-bonita.adoc
```diff
@@ -154,37 +154,40 @@ Make sure you double-check that you use the official driver version that matches
...
+|bpm
+|===
 
-  | db.password
-  | The password used to authenticate to the database
-  | bpm
-  |===
+[NOTE]
+====
+*If you are using MySQL, add `?allowMultiQueries=true` to the URL. +
```
alachambre, line 188: I believe this star is not here on purpose, I suggest to remove it, it feels weird.
```suggestion
If you are using MySQL, add `?allowMultiQueries=true` to the URL. +
```

alachambre (commented):
Looks good 👍 
Just the suggestion on the star, decide if you want to keep it or not, and then feel free to merge :)

modules/version-update/pages/migrate-from-an-earlier-version-of-bonita.adoc
NathalieC: This star recalls the star in the table, for the database URL row.

PR #1806, opened 2021-10-28
fix(docker): fix docker-compose file was not working

baptistemesta:


baptistemesta:
`2022.1` version need to change because it's not working with the removal of db-tool

modules/runtime/pages/bonita-docker-installation.adoc
```diff
@@ -168,25 +168,23 @@ services:
       - |
         set -e
         echo 'Waiting for Postgres to be available'
-        export PGPASSWORD="$$POSTGRES_ENV_POSTGRES_PASSWORD"
+        export PGPASSWORD=$${POSTGRES_ENV_POSTGRES_PASSWORD}
         maxTries=10
-        while [ "$$maxTries" -gt 0 ] && ! psql -h "$$DB_HOST" -U 'postgres' -c '\l'; do
+        while [[ maxTries -gt 0 ]] && ! psql -h $${DB_HOST} -U 'postgres' -c '\l'; do
```
passga: in 2021.2, the psql tool is no more installed in docker image so we could not use it to wait for that database to be available.

modules/runtime/pages/bonita-docker-installation.adoc
```diff
@@ -128,7 +128,7 @@ docker run --name bonita_postgres -h <hostname> -v bonita-lic:/opt/bonita_lic/ -
 
 ==== Using docker-compose
 
-Create a file stack.yml with the following content (remember to replace the `<hostname>` with the one used in the licence generation command):
+Create a file `docker-compose.yml` with the following content
 
 [source,shell script,subs="+macros"]
```
alachambre: It seems that the source code below is YML and not shell script. 
You should update it, else the comment on the first line doesn't looks like a comment.

modules/runtime/pages/bonita-docker-installation.adoc
baptistemesta: Yes, It's not available anymore in 2022.1, I planned to change it in that version

modules/runtime/pages/bonita-docker-installation.adoc
baptistemesta: Nice catch! thanks

modules/runtime/pages/bonita-docker-installation.adoc
passga: sorry  you right, i did mistake with version number :(

modules/runtime/pages/bonita-docker-installation.adoc
```diff
@@ -168,25 +168,23 @@ services:
...
+        export PGPASSWORD=$${POSTGRES_ENV_POSTGRES_PASSWORD}
         maxTries=10
-        while [ "$$maxTries" -gt 0 ] && ! psql -h "$$DB_HOST" -U 'postgres' -c '\l'; do
+        while [[ maxTries -gt 0 ]] && ! psql -h $${DB_HOST} -U 'postgres' -c '\l'; do
             let maxTries--
             sleep 1
         done
-        echo
-        if [ "$$maxTries" -le 0 ]; then
+        if [[ maxTries -le 0 ]]; then
```
passga: maxTries should be define as variable `$maxTries`, no ?

modules/runtime/pages/bonita-docker-installation.adoc
passga: maxTries should be define as variable $maxTries, no ?

modules/runtime/pages/bonita-docker-installation.adoc
baptistemesta: it's define above (172)

baptistemesta: I reverted that part: both `"$$maxTries"` and `maxTries` work. Not `$maxTries`

passga: ok

passga (approved):
LGTM!!

PR #1817, opened 2021-11-08
fix: add more info on case-sensitiveness in search

baptistemesta:
Relates to [RUNTIME-590](https://bonitasoft.atlassian.net/browse/RUNTIME-590)

### DIFF
--- a/modules/api/pages/using-list-and-search-methods.adoc
+++ b/modules/api/pages/using-list-and-search-methods.adoc
@@ -55,7 +55,7 @@ sob.differentFrom</b>(HumanTaskInstanceSearchDescriptor.ASSIGNEE_ID, myUser.getI
 [#case_sensitivity]
 == Case sensitivity
 
-WARNING: On Oracle database, search are case-sensitive.
+WARNING: On Oracle database, search is case-sensitive.
 
 All Search done using search terms are case-insensitive but not filter.
 
--- a/modules/api/pages/using-list-and-search-methods.adoc
+++ b/modules/api/pages/using-list-and-search-methods.adoc
@@ -79,7 +79,6 @@ All Search requests done using **search terms** are case-insensitive but **filte
 |===
 
 [#word_based_search]
-
 == Word-based search
 
 By default, search uses a "starts by" algorithm for searching, and finds matches where the search string occurs at the start of a phrase.
--- a/modules/api/pages/using-list-and-search-methods.adoc
+++ b/modules/api/pages/using-list-and-search-methods.adoc
@@ -57,7 +57,7 @@ sob.differentFrom</b>(HumanTaskInstanceSearchDescriptor.ASSIGNEE_ID, myUser.getI
 
 WARNING: On Oracle database, search is case-sensitive.
 
-All Search done using search terms are case-insensitive but not filter.
+All Search requests done using **search terms** are case-insensitive but **filtering** IS case-sensitive.
 
 |===
 | Phrase in database | Search string | Matches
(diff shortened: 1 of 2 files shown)


modules/api/pages/using-list-and-search-methods.adoc
```diff
@@ -51,6 +51,34 @@ sob.or();
 sob.differentFrom</b>(HumanTaskInstanceSearchDescriptor.ASSIGNEE_ID, myUser.getId());
 ----
 
+
+[#case_sensitivity]
+== Case sensitivity
+
+WARNING: On Oracle database, search are case-sensitive.
```
educhastenier:
```suggestion
WARNING: On Oracle database, search is case-sensitive.
```

modules/api/pages/using-list-and-search-methods.adoc
```diff
@@ -51,6 +51,34 @@ sob.or();
...
+
+All Search done using search terms are case-insensitive but not filter.
```
educhastenier:
```suggestion
All Search requests done using **search terms** are case-insensitive but **filtering** IS case-sensitive.
```

modules/api/pages/using-list-and-search-methods.adoc
```diff
@@ -51,6 +51,34 @@ sob.or();
...
+| Hello Charles
+| `searchBuilder.filter("name", "Hello Charles")`
+| yes
+
+| Hello Charles
+| `searchBuilder.filter("name", "hello charles")`
+| no
+
+| Hello Charles
+| `searchBuilder.searchTerm( "Hello Charles")`
```
educhastenier:
```suggestion
| `searchBuilder.searchTerm("Hello Charles")`
```

modules/api/pages/using-list-and-search-methods.adoc
```diff
@@ -51,6 +51,34 @@ sob.or();
 sob.differentFrom</b>(HumanTaskInstanceSearchDescriptor.ASSIGNEE_ID, myUser.getId());
 ----
 
+
+[#case_sensitivity]
```
alachambre: For the record you do not need to explicitly write anchors, they are generated automatically by Antora -> https://docs.asciidoctor.org/asciidoc/latest/macros/xref/#anchors

modules/api/pages/using-list-and-search-methods.adoc
```diff
@@ -51,6 +51,34 @@ sob.or();
...
+| Hello Charles
+| `searchBuilder.searchTerm("Hello Charles")`
+| yes
+
+| Hello Charles
+| `searchBuilder.searchTerm("hello charles")`
+| yes
+|===
+
 [#word_based_search]
```
alachambre, line 81:
```suggestion
```

modules/api/pages/using-list-and-search-methods.adoc
```diff
@@ -51,6 +51,34 @@ sob.or();
 sob.differentFrom</b>(HumanTaskInstanceSearchDescriptor.ASSIGNEE_ID, myUser.getId());
 ----
 
+
+[#case_sensitivity]
+== Case sensitivity
+
+WARNING: On Oracle database, search is case-sensitive.
+
+All Search requests done using **search terms** are case-insensitive but **filtering** IS case-sensitive.
```
alachambre, line 58: It might be because i'm not familiar with the subject, but I'm not sure to understand when I have case-sensitiveness and when I don't: 

- On the warning above, you say that 'On oracle, **search** is case sensitive'
- Then you say that search requests using search terms are case-insensitive but filtering is case sensitive

It's a bit confusing and definitively not straight forward, you start your explanation with an exception and  then you state a rule. 
I suggest you to remove the warning an to simply write two sentences: 
- On most of the database, ...
- On Orcale, ...

baptistemesta: I agree. I'll change that!

baptistemesta:
Merging that as feedback as been integrated

PR #1821, opened 2021-11-09
fix: remove table for non table block

alachambre:
Probable a conversion issue.


tbouffard (approved):
👍🏿 Nice catch

Before | After
-------- | --------
image | image

PR #1838, opened 2021-11-17
update(resource management): line breaks

NathalieC:
* add the note about line breaks in the page.properties file

### DIFF
--- a/modules/ROOT/pages/resource-management.adoc
+++ b/modules/ROOT/pages/resource-management.adoc
@@ -30,7 +30,7 @@ contentType=layout
 
 If a resource contains both `Index.groovy` and `index.html`, the Groovy class takes precedence.
 
-[NOTE]
+[TIP]
 ====
 To insert line breaks in a page.properties file, end the line that precedes the line break with a `\`.  +
 For example:
@@ -39,6 +39,7 @@ apiExtensions=example1RestAPI, example2RestAPI, example3RestAPI, example4RestAPI
 example5RestAPI, example6RestAPI, example7RestAPI, example8RestAPI, \
 example9RestAPI, example10RestAPI
 ```
+For more information on advanced properties file (line break, natural lines, logical lines...) please refer to the https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#load(java.io.Reader)[official documentation] 
 ====
 
 == Authorization permissions for resources
--- a/modules/ROOT/pages/resource-management.adoc
+++ b/modules/ROOT/pages/resource-management.adoc
@@ -30,6 +30,12 @@ contentType=layout
 
 If a resource contains both `Index.groovy` and `index.html`, the Groovy class takes precedence.
 
+[NOTE]
+
+====
+To insert line breaks in a page.properties file, end the line that preceeds the line break with a "\".
+====
+
 == Authorization permissions for resources
 
 If you are using xref:rest-api-authorization.adoc[REST API authorization] and your resource needs the Web REST API,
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/resource-management.adoc
```diff
@@ -30,6 +30,12 @@ contentType=layout
 
 If a resource contains both `Index.groovy` and `index.html`, the Groovy class takes precedence.
 
+[NOTE]
+
```
alachambre:
```suggestion
```

tbouffard: extra line break
```suggestion
```

modules/ROOT/pages/resource-management.adoc
```diff
@@ -30,6 +30,12 @@ contentType=layout
...
+====
+To insert line breaks in a page.properties file, end the line that preceeds the line break with a "\".
+====
```
tbouffard, line 43: I suggest to provide an example of properties source file in the admonition block. The user will be able to better understand what we mean.

modules/ROOT/pages/resource-management.adoc
```diff
@@ -30,6 +30,12 @@ contentType=layout
 
 If a resource contains both `Index.groovy` and `index.html`, the Groovy class takes precedence.
 
+[NOTE]
+
+====
+To insert line breaks in a page.properties file, end the line that preceeds the line break with a "\".
```
alachambre: I suggest you ton point to the official properties javadoc so users will have the fully detailed explanation, with all the important things to take into account to create 'advanced' properties file. 
I suggest you to put it in a TIP admonition, with a content like 

"To write advanced properties file (line break, natural lines, logical lines...) please refer to the official documentation -> https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#load(java.io.Reader)

alachambre (changes requested):

modules/ROOT/pages/resource-management.adoc
tbouffard: Typo + Remove the double quote: they must be not added in the file
```suggestion
To insert line breaks in a page.properties file, end the line that precedes the line break with a `\`.
```

tbouffard: Current rendering
image

alachambre: @NathalieC What about this pull request? 
If you don't want to take it account suggestions let us know.

modules/ROOT/pages/resource-management.adoc
NathalieC: ok. @mariellebonita ; can you help by providing an example following the case you have solved on this very topic? That would be great.

marielle-spiteri: example added

marielle-spiteri (commented):
example added

modules/ROOT/pages/resource-management.adoc
alachambre: @NathalieC  UUUUPP :)

alachambre:
@NathalieC Don't forget your pull request please :)

modules/ROOT/pages/resource-management.adoc
NathalieC: Here I am. Coming.

modules/ROOT/pages/resource-management.adoc
NathalieC: Thanks Marielle!

modules/ROOT/pages/resource-management.adoc
NathalieC: I've added what I think to be the best of both worlds: a quick tip for the use case we know is an issue (line breaks) + the official link to the documentation, to deepen one's knowledge on properties.

alachambre (approved):

PR #1843, opened 2021-11-23
Continue Improvements on the archetypes

florentsarat:
Harmonize the archetype documentation

PR #1844, opened 2021-11-25
fix(create a connector): wrong trailing char for first command and ex…

jjargot:
- wrong / trailing char replaced with \
- all properties specified on the command line example


rbioteau:
Can you remake the pr for previous versions (since 7.10)

florentsarat:
Closed as it covered by https://github.com/bonitasoft/bonita-doc/pull/1843

PR #1846, opened 2021-11-26
Update multi-tenancy-and-tenant-configuration.adoc

vbotnari12:
Multi-tenancy is deprecated at Bonita starting from version 2021.1, hence it should be reflected in the documentation with : 
- a warning message informing that the feature is deprecated but still supported
- an error message informing that the feature is not for free even for customers
- by removing the tenant creation procedure

For version 7.10 and 7.11 we have to display the warning and error but keep the procedure - > dedicated PR for it #1846

PR #1848, opened 2021-11-26
Update multi-tenancy-and-tenant-configuration.adoc

vbotnari12:
Multi-tenancy is deprecated at Bonita starting from version 2021.1, hence it should be reflected in the documentation with :

- a warning message informing that the feature is deprecated but still supported
- an error message informing that the feature is not for free even for customers

For version 2021.1 and 2021.2 the procedure has to be removed - > handling in PR #1846

### DIFF
--- a/modules/ROOT/pages/multi-tenancy-and-tenant-configuration.adoc
+++ b/modules/ROOT/pages/multi-tenancy-and-tenant-configuration.adoc
@@ -3,9 +3,14 @@
 
 One single platform to administrate with multiple logical engines. Discover what is a Tenant and how to setup Bonita with Multi-tenancy.
 
+[NOTE]
+====
+For Enterprise and Performance editions only.
+====
+
 [WARNING]
 ====
-This feature was deprecated in Bonita 2021.1. This means that Multi-tenant architecture will continue to work in Bonita for existing Bonita Enterprise and Performance subscription customers,
+This feature was deprecated in Bonita 2021.1. This means that Multi-tenant architecture will continue to work in Bonita for existing Bonita subscription customers,
 but it will not be offered to new customers, nor offered as a modification for existing Bonita subscription customers who are not already using multi-tenancy.
 ====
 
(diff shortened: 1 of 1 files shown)


tbouffard:
@vbotnari12 can you provide a convenient name for the Pull Request. It should explain what it is changing.

modules/ROOT/pages/multi-tenancy-and-tenant-configuration.adoc
```diff
@@ -9,6 +9,17 @@ One single platform to administrate with multiple logical engines. Discover what
...
+[WARNING]
+====
+This feature was deprecated in Bonita 2021.1. This means that Multi-tenant architecture will continue to work in Bonita for existing Bonita subscription customers,
+but it will not be offered to new customers, nor offered as a modification for existing Bonita subscription customers who are not already using multi-tenancy.
+====
+
+[IMPORTANT]
+====
+Bonita Enterprise customers need a contractual agreement to create additional tenants. For any new tenant creation, any general or contractual question related to Multi-tenancy, please contact your customer success representative.
+====
```
tbouffard, line 20: ⚠️ Using too many different kinds of admonitions make the documentation hard to read.
Admonitions are here to highlight something specific, but here, everything is highlighted. Too many highlights ~ no highlights.
https://bonitasoft-bonita-doc-build_preview-pr-1848.surge.sh/bonita/7.10/multi-tenancy-and-tenant-configuration

cc @alachambre @NathalieC 

image

vbotnari12: I have removed the "Note" but the information in "warning" and "error" should be displayed this way.

tbouffard: The information about the editions supporting the feature is no more consistent with the other pages: we  always display it in a NOTE admonition. We plan to manage it in a different way in the future and this part will be generated. This will avoid manually declaring the block everywhere. So doing it differently here will also introduce future side effects.

Having both a WARNING and an IMPORTANT block is still weird: why not putting a single IMPORTANT block?
The message you want to convey can fit in a single block
- this is deprecated and candidate for removal
- if you are still using it, talk to your CS representative

Here is what we currently see
image

vbotnari12: Because that's how CS wants to have the information displaied. We had a single warning paragraph and clients said it wasnt clear. Hence the changes. I'll add back the note and that will be my last change. Thank you for understanding :)

pabloalonso (approved):
LGTM

PR #1853, opened 2021-12-01
Update release-notes.adoc

NathalieC:
* reorder bugs from small to big numbers
* fix content within brackets
* add 2021.2 and 2021.2-u1 release date
* edit some bug titles to try to clarify their contents + name the issue,  not the fix

PR #1854, opened 2021-12-01
feat(product-versioning): provide more details about the releases

NathalieC:
* Add a first table of correspondance between Platform version, Technical Id and UID versions

### DIFF
--- a/modules/ROOT/pages/product-versioning.adoc
+++ b/modules/ROOT/pages/product-versioning.adoc
@@ -21,13 +21,50 @@ A technical Id also exists for Bonita components, and it follows the usual seman
 
 For example, upgrading from a 2021.1-0205 to a 2021.1-0401 will not require any migration. However, upgrades from a 7.11.5 to a 2021.1-0205 will require a migration.
 
-Technical Id Bonita version correspondence table
+Here is a Bonita Platform version / Technical Id / UI Designer version correspondence table:
 
 |===
-| Bonita Version | Technical Id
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/product-versioning.adoc
```diff
@@ -21,13 +21,50 @@ A technical Id also exists for Bonita components, and it follows the usual seman
 
 For example, upgrading from a 2021.1-0205 to a 2021.1-0401 will not require any migration. However, upgrades from a 7.11.5 to a 2021.1-0205 will require a migration.
 
-Technical Id Bonita version correspondence table
+Here is a Bonita Platform version / Technical Id / UI Designer version correspondence table:
 
 |===
-| Bonita Version | Technical Id
-
+| Release date | Bonita Platform version | Technical Id | UI Designer version
```
tbouffard, line 27: Are you sure you want to put release dates here? This adds maintenance tasks.
They are often missing in the release notes (see #1853 for instance), so I guess this will be the case here too.

Anyway, as this page must now be updated on new releases as the release-notes page, I guess we can enforce it is updated as part of the Release process, so adding the date is not a big deal.

NathalieC: Thanks for this feedback.
We will need to add the release numbers for each release, so I think the date is not that much to add on the top of it.
Yes, the newly-arrived POs are now aware that the date is needed for each release in the release notes as well.
And yes, this will be part of Serenity.

tbouffard (approved):
LGTM

PR #1859, opened 2021-12-07
fix(deleteDocumentContent): fix example

ugaston:
The processAPI.deleteContentOfArchivedDocument() method expects 
--> the id of the archived document: archiveDocument.getId() = id of element in ARCH_DOCUMENT_MAPPING table
--> and NOT the id of the document content: archiveDocument.contentStorageId = id of the document content in DOCUMENT table

Versions: 7.11, 2021.1, 2021.2


marielle-spiteri (approved):
LGTM

baptistemesta:
Good catch 👍

PR #1861, opened 2021-12-08
feat(datasource): Add perfomance tuning for bonita datasource

danila-m:
Relates to [RUNTIME-474](https://bonitasoft.atlassian.net/browse/RUNTIME-474)

### DIFF
--- a/modules/ROOT/pages/performance-tuning.adoc
+++ b/modules/ROOT/pages/performance-tuning.adoc
@@ -281,6 +281,20 @@ Two datasources are defined:
 
 Note that the sum of the maximum values configured for bonitaDS and bonitaSequenceManagerDS should be less than or equal to the maximum number of simultaneous connections allowed to your database.
 
+TIP: While the configuration of the 2 provided datasources works,
+some improvements can be made (and were added in later bonita versions)
+----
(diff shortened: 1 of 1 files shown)


baptistemesta:
Do we have a troubleshooting section somewhere. 
I guess the goal was to explain that if the platform admin encounters connection issue like it happened in the customer case,  he should apply that fix.
If you think it's not usefull to add it there because it's already in the product, it';s better to not add anything more in the doc on that.

modules/ROOT/pages/performance-tuning.adoc
```diff
@@ -281,6 +281,20 @@ Two datasources are defined:
 
 Note that the sum of the maximum values configured for bonitaDS and bonitaSequenceManagerDS should be less than or equal to the maximum number of simultaneous connections allowed to your database.
 
+TIP: While the configuration of the 2 provided datasources works,
```
alachambre, line 284: You TIP doesn't embed the code example, is it what you had in mind?

alachambre:
@danila-m don't forget your pull request please :)

modules/ROOT/pages/performance-tuning.adoc
educhastenier: Up @danila-m

PR #1874, opened 2021-12-30
Delete unused pages

NathalieC:
* delete deployment-index.adoc page as it was not called from anywhere. Will be recreated when time comes.


alachambre:
@NathalieC 25 days without activity, this is way too much, please do not forget your pull requests.

PR #1876, opened 2022-01-04
chore: add github workflow to check PR title and body

tbouffard:
We document what the title and body should be but this is not also followed.
This workflow adds a failed status to the PR to remind everybody to follow the
guidelines.

**Note**: the worfklow definition is copied from https://github.com/bonitasoft/bonita-documentation-site/blob/35e5eaa92578676a80ae7e946d9681b946f56b38/.github/workflows/commit-message-check-format.yml and is known to work.
See also https://github.com/bonitasoft/bonita-documentation-site/pull/289

### DIFF
--- a/.github/workflows/commit-message-check-format.yml
+++ b/.github/workflows/commit-message-check-format.yml
@@ -8,9 +8,6 @@ jobs:
   check-for-cc:
     runs-on: ubuntu-20.04
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
       - name: check-for-cc
         id: check-for-cc
         uses: agenthunt/conventional-commit-checker-action@v1.0.0
--- a/.github/workflows/commit-message-check-format.yml
+++ b/.github/workflows/commit-message-check-format.yml
@@ -1,0 +1,18 @@
+name: Commit Message format check
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
+
+jobs:
+  check-for-cc:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: check-for-cc
+        id: check-for-cc
+        uses: agenthunt/conventional-commit-checker-action@v1.0.0
+        with:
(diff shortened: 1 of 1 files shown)


alachambre:
The regex is useless, you just copied the default one -> https://github.com/marketplace/actions/conventional-commit-checker

alachambre:
Actually there is a tiny difference, is it voluntary? I think for the pr body we should stick with the conventional validation. (which is quite easy to pass).

tbouffard:
> Actually there is a tiny difference, is it voluntary? I think for the pr body we should stick with the conventional validation. (which is quite easy to pass).

We introduced it in february https://github.com/bonitasoft/bonita-documentation-site/pull/102. You validated the PR created by @benjaminParisel.

our regexp: `'(.*\n*)+(.*)'`
default: `(.*\n)+(.*)`

[UPDATE] our regexp allow an empty PR body.

Something we can improve: add a comment to explain why we need to override the default and what we change

alachambre:
I'm pretty sur both regex allow an empty body, the + is enough to give the possibility to skip the first group. 

But it's fine we cvan merge it this way, I just think the * on the line break is not that usefull.

tbouffard:
> I'm pretty sur both regex allow an empty body, the + is enough to give the possibility to skip the first group.
> 
> But it's fine we cvan merge it this way, I just think the * on the line break is not that usefull.

I have tested both regexp with https://regex101.com/ using the ECMAScript flavor and only our custom rexep work for empty body.
I don't have a clear opinion on the 2 options
- keep the default
- use the custom regexp with a comment explaining why we are using it

As you said, we can choose a solution and easily switch to the other if it doesn't fit our need.
@benjaminParisel can you decide for us? 😺

benjaminParisel:
I think it's better to autohorize empty body. 

When we push some fix or typo in documentation, it's not necessary to add body to explain what we do. 

Sorry @alachambre, i prefer @tbouffard proposal.

I will merge and we can update it later if we want/need.

alachambre:
Hum OK, but then this is not the best way to write the regex, we should simply use `(.*\n)*(.*)`

PR #1877, opened 2022-01-04
feat(2021.2): activate pagination links in all pages

tbouffard:
I propose to add the links to all pages starting from 2021.2 as in this version, the taxonomy has been reordered and the content refreshed for a more guided tour.

It is possible to apply the changes to 

- a previous version
- to a subset of pages (but this requires more configuration)
- to the whole documentation site

Please let me know by posting a comment.

closes https://github.com/bonitasoft/bonita-documentation-site/issues/273

### Initial proposal

Bottom of https://bonitasoft-bonita-doc-build_preview-pr-1877.surge.sh/bonita/2021.2/release-notes

**_light theme_**
image

**_dark theme_**
image

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -8,6 +8,7 @@ asciidoc:
     javadocVersion: 7.13
     crowdinBonitaVersion: 7.13
     # page attributes
+    page-pagination: true # display links to previous/next pages at the bottom of the page
     page-editable: true
     page-out-of-support: false
 nav:


alachambre:
owiiiiii

alachambre:
does it come with antora 3? I never noticed this feature before

tbouffard:
> does it come with antora 3? I never noticed this feature before

It is already available (no link with Antora 3) and comes from the Antora Default UI (the version we initially used from end 2020 or early 2021), and work with the dark theme without any change. See https://github.com/bonitasoft/bonita-documentation-site/issues/273 for more details.

antora.yml
```diff
@@ -8,6 +8,7 @@ asciidoc:
     javadocVersion: 7.13
     crowdinBonitaVersion: 7.13
     # page attributes
+    page-pagination: true # display links to previous/next pages at the bottom of the page
```
educhastenier, line 11: Is it available to be displayed at the top of the page as well?

tbouffard: Not with the current implementation, but it should be technically possible to add it in the future.
@NathalieC what do you think about it? I don't plan to do any change in the theme for now i.e. pagination will be for now only at the bottom or won't be added.

NathalieC: I think it's an interesting feature, although not sure people actually need it besides the Getting Started pages.
But if it holds no usability risk at all, let's go for it. A few questions to check that aspect:

- How does it display with very long page names for both previous and next pages? 
- How does it handle the transition between two modules?
- How does it handle the transition between the last page of a chapter and the index page of the next chapter?

Plus, it may help us realize our page names are not that great and need some improvements (like "Bonita" in the example you have provided :/)
 "it is possible to apply the changes to the whole documentation site" --> Yeah, I think this is what we should do.

NathalieC: okay, I've checked my 3 use-cases through the preview; it behaves great.
So, let's go for it :)

tbouffard:
OK, based on all positive feedback, I am going to merge now as it is.
This will let us use and learn with the latest Bonita versions.
Later, we will be able to decide if we apply it to older Bonita versions or even to other or all components.

PR #1878, opened 2022-01-07
feat(test-toolkit): Create documentation for test extension and users

alachambre:


modules/process-testing/pages/user.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+== Retrieve a User
+
+Users are mainly used to start processes and execute tasks. Any user in the organization can be retrieved using its username and the Bonita Test Toolkit: 
+
+[source, Java]
+----
+class ProcessIT {
+
+    @RegisterExtension
```
rbioteau: Prefer the `@BonitaTests` annotation

modules/process-testing/pages/bonita-test-extension.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+**- Target runtime URL**: The targeted Bonita runtime URL. It is used to perform HTTP requests to interact with this runtime.
+
+**- Delete process instances**: Whether all existing process instances (including archived instances and their documents) should be deleted before tests execution.
+
+**- Clear BDM**: Whether the BDM database of the targeted Bonita should be cleared before tests execution. All existing business data will be removed.
+
+[source, Java]
+----
+ @RegisterExtension <1>
```
rbioteau: Prefer the `@BonitaTests` annotation

You may document both way.

rbioteau (changes requested):

modules/process-testing/pages/bonita-test-extension.adoc
alachambre: Hum OK. I think I'll also document the way to configure test container in this page.

rbioteau: I don't think we should document testcontainers as the goal is to use it with BCD on  a provisionned environment

alachambre: As you wish, but in my view the best solution is to use test container and the builder and the deployer. It's sad to see that we do not want to provide our users the best tooling to use.

PR #1880, opened 2022-01-12
chore(cleanup): remove old wildfly refs

baptistemesta:
Closes [RUNTIME-217](https://bonitasoft.atlassian.net/browse/RUNTIME-217)

### DIFF
--- a/modules/ROOT/pages/licenses.adoc
+++ b/modules/ROOT/pages/licenses.adoc
@@ -55,7 +55,7 @@ The license file will be sent to you by email.
 
 == Install the license for the very first time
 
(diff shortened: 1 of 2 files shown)


tbouffard:
@baptistemesta RUNTIME-217 mentions 2021.1 and you currently propose the changes for 2021.2. Is it done on purpose? 😸 
We dropped the Wildfly support in 7.10, so we may also apply the changes to 7.10 instead.

baptistemesta:
Hello.
Yep, I rebased it on 7.10

educhastenier:
I thought I had removed the latest forgotten references to WildFly in commit https://github.com/bonitasoft/bonita-doc/commit/3e25d68372998bbf26f824681e386d5d0114e677
But obviously not...

PR #1881, opened 2022-01-13
feat(application): explain how to create a custom process manager application

DumitruCorini:
covers [RUNTIME-468](https://bonitasoft.atlassian.net/browse/RUNTIME-468)

### DIFF
--- a/modules/runtime/pages/process-manager-application.adoc
+++ b/modules/runtime/pages/process-manager-application.adoc
@@ -3,7 +3,6 @@
 
 [NOTE]
 ====
-
 For Subscription editions only. +
 If your Bonita Subscription has been purchased before 2021, this Add-on can be purchased on top of your subscription.
 If you have purchased or renewed your Bonita Subscription in 2021 or after, BCD is fully part of your subscription.
--- a/modules/runtime/pages/process-manager-application.adoc
+++ b/modules/runtime/pages/process-manager-application.adoc
@@ -1,0 +1,61 @@
+= Creating a process manager application
+:description: This page explains how to create your own custom process manager application
+
+[NOTE]
+====
+
+For Subscription editions only. +
+If your Bonita Subscription has been purchased before 2021, this Add-on can be purchased on top of your subscription.
+If you have purchased or renewed your Bonita Subscription in 2021 or after, BCD is fully part of your subscription.
+====
+
+This page explains how to create your own custom process manager application. +
+This application functions like the Administrator application, but has a smaller scope, one that only controls processes, cases and tasks. +
+Creating such an application follows the same steps as the creation of any xref:applications:application-creation.adoc[other custom application].
+
+== Pages
+
+The main menu of the application will require four pages.
+
+* Monitoring
+* Process list
+* Case list
+* Task list
+
+In addition to those, there are some orphan pages that are required by the above ones :
+
+* Process details
+* Process visualization
+* Case details
+* Case visualization
+* Task details
+
+== Page modifications
+
+The above-mentioned menu pages are the same as the ones that are part of the Administrator application, but they only show the processes/cases/tasks that the user supervises. +
+In order to make the menu pages do so, you need to export them from the Administrator Resource List and edit them using the UI Designer.
+You don't need to export and modify the orphan pages.
+
+[WARNING]
+====
+
+Currently, the Monitoring page is made using a different technology than UI Designer, and thus it cannot be edited.
+====
+
+The modifications that you will need to do are quite simple:
+
+1. Though not necessary, I highly recommend re-saving your pages with a different name. This will prevent the pages from being over-written if you need to re-import the default pages.
+2. Add an api call to get current user session. Create a new "External API" variable called `session` with the value `../API/system/session/unusedId`
+3. Add `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).
+
+== Application creation
+
+After all the menu pages have been modified, we can proceed with the application creation. We can create the application from scratch, but it's easier to : +
+
+1. Export the Bonita Administrator Application from the Application List and import it into the Bonita Studio
+2. Delete any orphan page that isn't mentioned in the list above
+3. Delete all, but the BPM menu from the "Navigation" section
+4. Create application pages in the "Navigation" menu using the menu pages that we mentioned above. Make sure to keep the tokens of the pages as the ones from the Administrator application, since the pages make use of the specific tokens or, if you modify the tokens, modify the pages to also use the correct urls.
+5. Rename your application and modify the application token
+
+After doing all of those, you can deploy the finished "Process manager" application.
--- a/modules/runtime/pages/process-manager-application.adoc
+++ b/modules/runtime/pages/process-manager-application.adoc
@@ -1,4 +1,4 @@
-= Creating a process manager application
+= Create a process manager application
 :description: This page explains how to create your own custom process manager application
 
 [NOTE]
@@ -19,7 +19,7 @@ The main menu of the application will require four pages.
 * Case list
 * Task list
 
-In addition to those, there are some orphan pages that are required by the above ones :
+In addition to those, some orphan pages are required by the above ones :
 
 * Process details
 * Process visualization
@@ -39,7 +39,7 @@ You don't need to export and modify the orphan pages.
 Currently, the Monitoring page is made using a different technology than UI Designer, and thus it cannot be edited.
 ====
 
-The modifications that you will need to do are quite simple:
+Perform the following modifications to the 4 menu pages:
 
 1. Though not necessary, it is highly recommended to re-save your pages with a different name. This will prevent these pages from being over-written if you need to re-import the default pages.
 2. Add an api call to get current user session. Create a new "External API" variable called `session` with the value `../API/system/session/unusedId`
--- a/modules/runtime/pages/process-manager-application.adoc
+++ b/modules/runtime/pages/process-manager-application.adoc
@@ -3,9 +3,7 @@
 
 [NOTE]
 ====
-For Subscription editions only. +
-If your Bonita Subscription has been purchased before 2021, this Add-on can be purchased on top of your subscription.
-If you have purchased or renewed your Bonita Subscription in 2021 or after, BCD is fully part of your subscription.
+For Enterprise, Performance, Efficiency and Teamwork editions only.
 ====
 
 This page explains how to create your own custom process manager application. +
@@ -43,18 +41,19 @@ Currently, the Monitoring page is made using a different technology than UI Desi
 
 The modifications that you will need to do are quite simple:
 
-1. Though not necessary, I highly recommend re-saving your pages with a different name. This will prevent the pages from being over-written if you need to re-import the default pages.
+1. Though not necessary, it is highly recommended to re-save your pages with a different name. This will prevent these pages from being over-written if you need to re-import the default pages.
 2. Add an api call to get current user session. Create a new "External API" variable called `session` with the value `../API/system/session/unusedId`
-3. Add `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).
+3. Make all API calls provide only the processes/cases/tasks that he supervises by adding `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).
 
+[#application-creation]
 == Application creation
 
-After all the menu pages have been modified, we can proceed with the application creation. We can create the application from scratch, but it's easier to : +
+After all the menu pages have been modified, the application can be created. To do so : +
 
 1. Export the Bonita Administrator Application from the Application List and import it into the Bonita Studio
 2. Delete any orphan page that isn't mentioned in the list above
 3. Delete all, but the BPM menu from the "Navigation" section
-4. Create application pages in the "Navigation" menu using the menu pages that we mentioned above. Make sure to keep the tokens of the pages as the ones from the Administrator application, since the pages make use of the specific tokens or, if you modify the tokens, modify the pages to also use the correct urls.
+4. Create application pages in the "Navigation" menu using the menu pages that we mentioned above. It is highly recommended to not modify the tokens of the pages, since the pages use of the specific tokens. If you modify the tokens, make sure to modify the pages to also use the correct urls.
 5. Rename your application and modify the application token
 
-After doing all of those, you can deploy the finished "Process manager" application.
+The "Process manager" application is now complete, ready to be deployed and used.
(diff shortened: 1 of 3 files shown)


modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,61 @@
+= Creating a process manager application
+:description: This page explains how to create your own custom process manager application
+
+[NOTE]
+====
+
+For Subscription editions only. +
```
tbouffard: ⚠️ this is not the wording we use in other pages. We always reference Teamwork, ..., Enterprise.
Please check with @NathalieC 
Not using consistent messaging is a pain for readers and will make maintenance harder. Please see https://github.com/bonitasoft/bonita-documentation-site/issues/279

DumitruCorini: Thank you very much for your comment. I'll look into it

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,61 @@
+= Creating a process manager application
+:description: This page explains how to create your own custom process manager application
+
+[NOTE]
+====
+
```
alachambre: remove useless blank line
```suggestion
```

modules/runtime/pages/process-manager-application.adoc
NathalieC: Hello. I have no statistics to provide on this topic, but a quick search brings back a little bit of both.
Here is why:
- When we have started with such notes, this was because we needed to explain a feature was available in some editions but not all, like Efficiency and Performance, but no Teamwork, or only in Performance, and not in Teamwork of Efficiency.
- When we have started to sell Enterprise editions and no more TW, Eff, Perf, the _all inclusive_ "For Subscriptions editions only" has come up as a handy way to express "teamwork, efficiency, performance, enterprise" as there was no need to differentiate between all 4, and it started to become looong to write. 
- But at the same time, on some pages, we have continued to write all 5 subscription editions.

**True** that we need to clarify.

So, my take on this is that:
- when a feature is about all Subscription editions, we should be short and say "For Subscription editions only".  I find it more effective, and I think this is a clear enough vocabulary for a user of any subscription edition.

- when a feature is for only some subscription editions but not all, we should write the details.

@tbouffard how does that sound to you?
If you agree, I can also modify the chapters referred to in https://github.com/bonitasoft/bonita-documentation-site/issues/279.
And I will take this item to check and replace when needed.

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
+= Creating a process manager application
+:description: This page explains how to create your own custom process manager application
+
+[NOTE]
+====
+For Subscription editions only. +
+If your Bonita Subscription has been purchased before 2021, this Add-on can be purchased on top of your subscription.
```
alachambre: Why do you talk about BCD here?? This page doesn't have anything to do with BCD.

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+[WARNING]
+====
+
+Currently, the Monitoring page is made using a different technology than UI Designer, and thus it cannot be edited.
+====
+
+The modifications that you will need to do are quite simple:
+
+1. Though not necessary, I highly recommend re-saving your pages with a different name. This will prevent the pages from being over-written if you need to re-import the default pages.
```
alachambre: use the impersonal form as much as possible in documentation, especially avoid the 'I'. Users do not want 'your' advice but 'the' recommendation. 
```suggestion
1. Though not necessary, it is highly recommended to re-save pages with a different name. This will prevent these pages from being over-written if you need to re-import the default pages.
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
...
+2. Add an api call to get current user session. Create a new "External API" variable called `session` with the value `../API/system/session/unusedId`
+3. Add `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).
```
alachambre: This is not understandable. 

People are not bots, explain why they have to do something and the code to write will become understandable.

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+== Application creation
+
+After all the menu pages have been modified, we can proceed with the application creation. We can create the application from scratch, but it's easier to : +
```
alachambre: Again, prefer the impersonal form. And never use 'easy' in a technical documentation, you do not know the level of your reader.

```suggestion
After all the menu pages have been modified, the application can be created. To do so : +
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+1. Export the Bonita Administrator Application from the Application List and import it into the Bonita Studio
+2. Delete any orphan page that isn't mentioned in the list above
```
alachambre: You could create an anchor for this list and add a link here. 
Today the paragraphs are closed, but you do not know how it will look tomorrow.

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
...
+3. Delete all, but the BPM menu from the "Navigation" section
+4. Create application pages in the "Navigation" menu using the menu pages that we mentioned above. Make sure to keep the tokens of the pages as the ones from the Administrator application, since the pages make use of the specific tokens or, if you modify the tokens, modify the pages to also use the correct urls.
```
alachambre: This sentence is too long, it's hard to read. 
And you start a sentence with 'make sure to..' for something that is recommended, but according to the end of the sentence not mandatory, it feels weird, I suggest to say it differently.

```suggestion
4. Create application pages in the "Navigation" menu using the menu pages mentioned above. It is highly recommended to keep the tokens of the pages as the ones from the Administrator application, since the pages make use of the specific tokens. If you modify these tokens, make sure to modify the pages to also use the correct urls.
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,60 @@
...
+5. Rename your application and modify the application token
+
+After doing all of those, you can deploy the finished "Process manager" application.
```
alachambre: I do not like this conclusion, but I don't really know why ^^ 
Here is a suggestion that looks better in my view. 
```suggestion
The "Process manager" application is now complete, ready to be deployed and used.
```

alachambre (changes requested):
Few suggestions, some modifications have to be done IMHO.

alachambre:
This new page has not been added in the navigation, I think it should be.

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: I think it was a copy paste from the first result that I got, my mistake

alachambre: Just put 'For Enterprise, Performance, Efficiency and Teamwork  editions only.' ;)

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: I replaced it with "Make all API calls provide only the processes/cases/tasks that he supervises by adding `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).". 
Please, human, tell me if it's more understandable that way 🤖

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: I'm just a bit confused about the creation of the anchor, I don't quite get why we want to create it

DumitruCorini:
I added your suggestions. Thanks a lot @alachambre @tbouffard @NathalieC. Feel free to add some other ones if you find something to improve.

modules/runtime/pages/process-manager-application.adoc
tbouffard: OK this makes sense.
But I am just afraid that introducing this new wording without explanation will make this harder to understand. Remember there is no explanation in the documentation nor on the website about the extra edition names.

Here is the list of `Editions` names we now have in the documentation:
- Community
- Teamwork
- Efficiency
- Performance
- Enterprise
- `Subscription editions` (the one you refer to)

About changing the other pages, I suggest we **decide how** to do it directly in https://github.com/bonitasoft/bonita-documentation-site/issues/279 and continue the discussion there. We have ways to reduce duplicating the message.

tbouffard: Notice there are places where we will continue to use the old editions, and this will probably the sole pages where the Teamwork name will be used 😿 

Like in https://documentation.bonitasoft.com/bonita/2021.2/identity/organization-management-in-bonita-bpm-studio#_profile_management

image

tbouffard:
@DumitruCorini I have nothing to add, I let others review the new content 😸.
Thanks for the discussions and having taken the whole review feedback into consideration.

modules/runtime/runtime-nav.adoc
```diff
@@ -37,6 +37,7 @@
   *** xref:administrator-application-index.adoc[Bonita Administrator application]
    **** xref:admin-application-overview.adoc[Overview]
    **** xref:process-management-index.adoc[Process Management]
+    ***** xref:process-manager-application.adoc[Creating a process manager application]
```
alachambre:
```suggestion
    ***** xref:process-manager-application.adoc[Create a process manager application]
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,59 @@
+= Creating a process manager application
```
alachambre: I think it's better to use 'Create' that 'Creating' in a title, especially in this context: we explain how to create a pm application, not what to do when you are creating a pm application (the difference is a bit subtil but still).

```suggestion
= Create a process manager application
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,59 @@
...
+== Pages
+
+The main menu of the application will require four pages.
+
+* Monitoring
+* Process list
+* Case list
+* Task list
+
+In addition to those, there are some orphan pages that are required by the above ones :
```
alachambre: This sentence isn't fluent, keep it simple 👍 

```suggestion
In addition to those, some orphan pages are required by the above ones :
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,59 @@
...
+In order to make the menu pages do so, you need to export them from the Administrator Resource List and edit them using the UI Designer.
+You don't need to export and modify the orphan pages.
+
+[WARNING]
+====
+
+Currently, the Monitoring page is made using a different technology than UI Designer, and thus it cannot be edited.
+====
+
+The modifications that you will need to do are quite simple:
```
alachambre: I do not recommend to qualify the edition of a complex page with the UID as something quite simple, it is probably the case for you, one of the dev of the original page, but i'm pretty sure that many users will find this complicated. 
As always, i suggest to avoid adjectives like 'easy, simple...', you do not know the level of your reader.

An other point is that i suggest you to be a bit more explicit about which pages have to be modified, I had to read the text 3 times to get it.

```suggestion
Perform the following modifications to the 4 menu pages:
```

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+1. Though not necessary, it is highly recommended to re-save your pages with a different name. This will prevent these pages from being over-written if you need to re-import the default pages.
+2. Add an api call to get current user session. Create a new "External API" variable called `session` with the value `../API/system/session/unusedId`
+3. Make all API calls provide only the processes/cases/tasks that he supervises by adding `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).
```
alachambre: It's still complicated to understand.  
Why are you afraid to make a real sentence with more than 3 words to explain yourself? 

```suggestion
3. Since a process manager only supervise a subpart of all the processes/cases/tasks, some API call must be updated to only retrieve the expected processes/cases/tasks. To do so, the filter `supervisor_id` has to be used with the current user ID: Add  `&f=supervisor_id={{session.user_id}}` to the variables of the "External API" type (except for the `session` variable that we created).
```

modules/runtime/pages/process-manager-application.adoc
alachambre: You reference a paragraphe of your page by saying 'the list above'. 
If tomorrow this page is updated, and a lot of text is added between this sentence and the list it will become unclear. 
By creating an anchor and adding a link to this anchor on the 'the list', you ensure that no matter what happen it should stay usable.

alachambre (changes requested):
Still a few things to change, and a few new suggestions.

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,59 @@
...
+* Task list
+
+In addition to those, there are some orphan pages that are required by the above ones :
+
+* Process details
+* Process visualization
+* Case details
+* Case visualization
+* Task details
+
```
cleosurville, line 30: Il n'y a pas les pages de vizualisation ou elles ne sont pas appelées comment ça ou je ne les aies pas trouvé.

modules/runtime/pages/process-manager-application.adoc
```diff
@@ -0,0 +1,59 @@
...
+The main menu of the application will require four pages.
+
+* Monitoring
+* Process list
+* Case list
+* Task list
+
+In addition to those, there are some orphan pages that are required by the above ones :
+
+* Process details
```
cleosurville, line 25: Impossible d'importer la page page admin process details

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: Ok, I understood now, I'll modify it

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: T'as pas besoin de l'importer, elle est deja presente dans les apps

DumitruCorini: Mais, ca veux dire que j'ai besoin de l'expliquer mieux

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: Done

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: I rephrased this a bit

modules/runtime/pages/process-manager-application.adoc
DumitruCorini: Ouais, mais t'as pas besoin de les importer

alachambre (approved):

PR #1883, opened 2022-01-14
chore(pr): Adding pr diff check

benjaminParisel:
* Only failed if https://documentation.bonitasoft.com/ is put in adoc content


tbouffard (approved):
Nice! 👍🏿

PR #1885, opened 2022-01-17
feat(logger): add info on how to activate old sysout log

baptistemesta:
In 2022.1, we will not log by default in system out. To revert back to
the old behavior there is a system property that can be set as explained
here.

Relates to
[RUNTIME-795](https://bonitasoft.atlassian.net/browse/RUNTIME-795)


alachambre (approved):

educhastenier (approved):
Looks good to me :+1:

PR #1889, opened 2022-01-18
feat(dyn permissions): update VAR name to use to de-activate them

educhastenier:
in docker image

### DIFF
--- a/modules/ROOT/pages/release-notes.adoc
+++ b/modules/ROOT/pages/release-notes.adoc
@@ -26,7 +26,7 @@ These credentials can be consulted and modified:
 
 === Dynamic authorization checking active by default (Subscription editions)
 
-xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is now *active by default* for new installations. +
+xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is*active by default* for new installations. +
 Existing installations that update to 2022.1 stay *unchanged*, but we encourage customers to activate dynamic authorization checking, once they have benched that the performance cost is reasonable.
 
 Community users that update to 2022.1 simply lose the possibility to use dynamic authorization checking.
--- a/modules/ROOT/pages/release-notes.adoc
+++ b/modules/ROOT/pages/release-notes.adoc
@@ -26,12 +26,12 @@ These credentials can be consulted and modified:
 
 === Dynamic authorization checking active by default (Subscription editions)
 
-xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is*active by default* for new installations. +
+xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is also *active by default* for new installations. +
 Existing installations that update to 2022.1 stay *unchanged*, but we encourage customers to activate dynamic authorization checking, once they have benched that the performance cost is reasonable.
 
-Community users that update to 2022.1 simply lose the possibility to use dynamic authorization checking.
-Their potential previous configuration regarding dynamic authorizations is kept unchanged during the update procedure,
-to be used as a reference for custom usage, or ready to use for a potential upgrade to Enterprise edition.
+For Community users who update to 2022.1, dynamic authorization checking does not work anymore. +
+However, the configuration stays unchanged, even if it is not used anymore. +
+If they want to reuse the configuration in a custom manner, or upgrade to Enterprise edition, the configuration will be back there and fully functional.
 
 == Technical updates
 
--- a/modules/ROOT/pages/release-notes.adoc
+++ b/modules/ROOT/pages/release-notes.adoc
@@ -23,6 +23,14 @@ These credentials can be consulted and modified:
 * in the <BONITA>/server/conf/tomcat-users.xml file in a *Bonita Tomcat bundle*
 * using xref:bonita-docker-installation.adoc#_monitoring_username[specific Environment parameters] when starting a *Bonita Docker image*
 
+
+=== Dynamic authorization checking active by default (Subscription editions)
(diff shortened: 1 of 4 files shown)


modules/ROOT/pages/release-notes.adoc
```diff
@@ -23,6 +23,14 @@ These credentials can be consulted and modified:
 * in the <BONITA>/server/conf/tomcat-users.xml file in a *Bonita Tomcat bundle*
 * using xref:bonita-docker-installation.adoc#_monitoring_username[specific Environment parameters] when starting a *Bonita Docker image*
 
+
+=== Dynamic authorization checking active by default (Subscription editions)
+
+xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is now *active by default* for new installations. +
+Existing installations that update to 2022.1 stay *unchanged*, but we encourage customers to activate dynamic authorization checking, once they have benched that the performance cost is reasonable.
+
+Community users that update to 2022.1 simply lose the possibility to use dynamic authorization checking, and their potential previous configuration regarding dynamic authorizations is automatically cleared out from the installation, during the update procedure.
```
educhastenier: This § is to be discussed

educhastenier: Will be discused next week in a meeing.
Putting this PR as WIP for now.

modules/ROOT/pages/release-notes.adoc
```diff
@@ -23,6 +23,14 @@ These credentials can be consulted and modified:
 * in the <BONITA>/server/conf/tomcat-users.xml file in a *Bonita Tomcat bundle*
 * using xref:bonita-docker-installation.adoc#_monitoring_username[specific Environment parameters] when starting a *Bonita Docker image*
 
+
+=== Dynamic authorization checking active by default (Subscription editions)
+
+xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is now *active by default* for new installations. +
```
alachambre: You use 'is now' twice in a few words, i suggest to find an other way to say it to avoid this repetition. 

Maybe just remove the second now? : 
```suggestion
xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is*active by default* for new installations. +
```

modules/ROOT/pages/release-notes.adoc
alachambre: Looks like a huge breaking change, I think it should be highlighted correctly to ensure that everybody get the message..

educhastenier: fixed

modules/ROOT/pages/release-notes.adoc
educhastenier: True
Thanks

modules/ROOT/pages/release-notes.adoc
educhastenier: Finally, we changed that behaviour.
I updated the doc accordingly.
Thx for the feedback @alachambre

modules/identity/pages/rest-api-authorization.adoc
```diff
@@ -395,30 +395,15 @@ edit `web.xml` and comment out the following definitions:
...
+If debug mode is activated, Bonita will reload the dynamic check scripts each time they are executed. It allows to rapidly validate your script at *development* time.
 
-To deactivate authorization, set the property to `false`, then restart the application server. Debug mode should be deactivated in production, so as not to impact performance.
+To activate debug mode, edit `bonita-platform-sp.properties` and set the value of the `bonita.runtime.authorization.dynamic-check.debug` property to `true`, then restart the application server.
 
-[#migrate]
+Then, each time you change a dynamic check script, simply update it in database using xref:runtime:bonita-platform-setup.adoc#update_platform_conf[Setup tool] (`./setup.sh push`). The script will be reloaded at next execution (next time you call a URL that matches this dynamic rule).
 
-[discrete]
-==== Migration
```
educhastenier, line 407: @abirembaut I chose to remove this § as it concerns very old versions.
Is it ok for you?

modules/ROOT/pages/release-notes.adoc
```diff
@@ -23,6 +23,16 @@ These credentials can be consulted and modified:
...
 
+
+=== Dynamic authorization checking active by default (Subscription editions)
+
+xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is*active by default* for new installations. +
+Existing installations that update to 2022.1 stay *unchanged*, but we encourage customers to activate dynamic authorization checking, once they have benched that the performance cost is reasonable.
+
+Community users that update to 2022.1 simply lose the possibility to use dynamic authorization checking.
+Their potential previous configuration regarding dynamic authorizations is kept unchanged during the update procedure,
+to be used as a reference for custom usage, or ready to use for a potential upgrade to Enterprise edition.
```
alachambre: As a reader who do not know the subject I'm not sure to understand, is the previous configuration still working or not? 
I suggest to be a bit more direct, if the conf is kept but doesn't work anymore say it, even if it is a bit redundant with the sentence 'lose the possibility to use dynamic authorization checking'. And if the conf is still working but community users cannot create new ones then some sentences should be changed.

educhastenier: The previous configuration is simply kept as is, but not taken into account anymore in Community edition.

@alachambre how do you suggest to rephrase it?

educhastenier: Here is another suggestion:
"For Community users who update to 2022.1,  dynamic authorization checking does not work anymore. +
However, the configuration stays unchanged, even if it is not used anymore. +
If they want to reuse the configuration in a custom manner, or upgrade to Enterprise edition, the configuration will be back there and fully functional."

What do you think?

alachambre: looks good 👍

educhastenier: integrated

PR #1897, opened 2022-01-25
feat(retry): update the retry mechanism description.

danila-m:
Following it's last modification in the engine, the retry mechanism behavior changed.

Relates to [RUNTIME-302](https://bonitasoft.atlassian.net/browse/RUNTIME-302)

### DIFF
--- a/modules/runtime/pages/fault-tolerance.adoc
+++ b/modules/runtime/pages/fault-tolerance.adoc
@@ -22,7 +22,7 @@ It is a reactive system, when an incident happens during the execution of a work
 analyses the exception and determine if the work should be retried automatically or not.
 
 When the exception is retryable, the platform queues the work again, with a delay. That can be done up to 10 times
-with the delay increasing gradually from 1 second to more than 1 hour by default.
+with the delay increasing gradually from 1 second to more than 1 hour by default. +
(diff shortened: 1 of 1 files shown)


danila-m:
Relates to : https://github.com/bonitasoft/bonita-engine-sp/pull/2161

modules/runtime/pages/fault-tolerance.adoc
```diff
@@ -23,6 +23,7 @@ analyses the exception and determine if the work should be retried automatically
 
 When the exception is retryable, the platform queues the work again, with a delay. That can be done up to 10 times
 with the delay increasing gradually from 1 second to more than 1 hour by default.
+The exact delay is randomized, to avoid repeated congestion situations (a typical case of those are database deadlocks, when retrying all the failed works at the same time will lead to more deadlocks).
```
educhastenier: Add dump carriage return:
```suggestion
with the delay increasing gradually from 1 second to more than 1 hour by default. +
The exact delay is randomized, to avoid repeated congestion situations (a typical case of those are database deadlocks, when retrying all the failed works at the same time will lead to more deadlocks).
```

educhastenier (commented):
otherwise, LGTM

PR #1898, opened 2022-01-26
feat(connectors): add doc how connector failure

baptistemesta:
Also explain that it's better to have idempotent connectors.

Closes [RUNTIME-767](https://bonitasoft.atlassian.net/browse/RUNTIME-767)

### DIFF
--- a/modules/api/pages/bpm-api.adoc
+++ b/modules/api/pages/bpm-api.adoc
@@ -4553,6 +4553,7 @@ Mandatory filters: connector_process_id, connector_name, connector_version
 ]
 ----
 
+[#connector-failure]
 === ConnectorFailure
 
 ==== Description
--- a/modules/api/pages/bpm-api.adoc
+++ b/modules/api/pages/bpm-api.adoc
@@ -4553,7 +4553,6 @@ Mandatory filters: connector_process_id, connector_name, connector_version
 ]
 ----
 
-[#connector-failure]
 === ConnectorFailure
 
 ==== Description
--- a/modules/process/pages/connector-archetype.adoc
+++ b/modules/process/pages/connector-archetype.adoc
@@ -310,6 +310,16 @@ class Connector extends AbstractConnector {
 The methods _validateInputParameters_ and _executeBusinessLogic_ must be implemented, and are called by the Bonita engine when the connector is executed. +
 The methods _connect_ and _disconnect_ can be used to open and close a connection to a remote server.  The life cycle of the connection will then be managed by the Bonita Runtime.
 
+
+[NOTE]
+====
+If possible, make the connector implementation idempotent. Subsequent execution of the same connector instance should
+not have additional side effects. This ensures, in rare case of failure when the engine is unable
+to determine if the connector was executed or not, that not unwanted side effects are produced.
+
+See xref:runtime:connectors-execution.adoc[connectors execution] for more information on how failures are handled on connectors.
+====
+
 ==== Add multiple implementations
 
 To add a new implementation create a `.impl` file in the `src/main/resources-filtered/` folder. They will be added to the `all` zip archive at build time. It is recommended to use the maven properties for implementation id and version.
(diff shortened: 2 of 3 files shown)


baptistemesta:
I've did it on `2021.2` because the page `connector-archetype.adoc` was not present before

modules/api/pages/bpm-api.adoc
```diff
@@ -4553,6 +4553,7 @@ Mandatory filters: connector_process_id, connector_name, connector_version
 ]
 ----
 
+[#connector-failure]
```
alachambre: For you information you do not have to manually create anchors for title, a default one is generated automatically -> https://docs.asciidoctor.org/asciidoc/latest/macros/xref/#anchors

modules/process/pages/connector-archetype.adoc
```diff
@@ -310,6 +310,16 @@ class Connector extends AbstractConnector {
 The methods _validateInputParameters_ and _executeBusinessLogic_ must be implemented, and are called by the Bonita engine when the connector is executed. +
 The methods _connect_ and _disconnect_ can be used to open and close a connection to a remote server.  The life cycle of the connection will then be managed by the Bonita Runtime.
 
+
+[NOTE]
+====
+If possible, make the connector implementation idempotent. Subsequent execution of the same connector instance should
```
alachambre: I think plural is more adequate here no?

```suggestion
If possible, make the connector implementation idempotent. Subsequent executions of the same connector instance should
```

modules/runtime/pages/connectors-execution.adoc
```diff
@@ -31,6 +31,21 @@ A typical example of this is if the custom code in a groovy connector falls into
 
 image::images/connector_execution_timeout.png[Connector execution with timeout]
 
+=== Connector failure handling
+
+When a connector fails, it is put as failed and the exception can be retrieved
+using the xref:api:bpm-api.adoc#connector-failure[dedicated api].
+
+Then the connector can be manually replayed using the xref:api:bpm-api.adoc#activity-replay[activityReplay api].
```
alachambre:
```suggestion
Then the connector can be manually be replayed using the xref:api:bpm-api.adoc#activity-replay[activityReplay api].
```

modules/runtime/pages/connectors-execution.adoc
```diff
@@ -31,6 +31,21 @@ A typical example of this is if the custom code in a groovy connector falls into
...
+
+
+[NOTE]
+====
+On rare case, when an error occurs on the connector, and the engine is unable to mark it as failed
```
alachambre:
```suggestion
On rare cases, when an error occurs on a connector and the engine is unable to mark it as failed
```

modules/runtime/pages/connectors-execution.adoc
```diff
@@ -31,6 +31,21 @@ A typical example of this is if the custom code in a groovy connector falls into
...
+(might be caused by the engine be unable to get a new sql connection) , the connector will automatically be retried.
```
alachambre:
```suggestion
(this might be caused by the engine being unable to get a new sql connection), the connector will automatically be retried.
```

alachambre (commented):
A few suggestions. 
And you should try to remember my point about auto-generated anchors, it's not the first time I made it :p

modules/api/pages/bpm-api.adoc
baptistemesta: I know but IntelliJ tells me it's better to had explicit anchors when referring to them. What is the best practice on that point?

alachambre: I don't really know what could be the best practice, IMHO when the generated anchor is short (i.e the title is short) use it, it avoids to add code.
If the title is too long then creating a custom anchor can be nice since the anchor appears in the final URL.

baptistemesta: Is there something that checks for broken links?
One. benefit of putting a that here is also to remember that we have a link and if we rename it we broke something. Otherwise, yep it's only noise.
I'll remove it

baptistemesta:
```suggestion
```

PR #1904, opened 2022-01-27
doc(REST): troubleshooting section for the content-range header

abirembaut:
* add troubleshooting section for the 416 error when the content-range header is blocked by some firewalls

Covers [RUNTIME-821](https://bonitasoft.atlassian.net/browse/RUNTIME-821)


alachambre:
I wonder if the title level of the section 'Troubleshooting' is correct in this page. 

At the moment it is a title lvl 4 under a big section 'getting started', but the content of the troubleshooting section seems to be general, not only about the getting started. 

I think it should be a section lvl 2, and not a lvl 4. It would make more sens in my view, and this way it will appear in the table of content and will be better indexed in the search.

abirembaut:
ok. I will look into it

Shmayro (approved):

PR #1909, opened 2022-01-31
chore(7.10): mark as 'out of support'

tbouffard:
7.10 is no more supported, so content should not be edited anymore. So mark pages as 'not editable'.
The configuration should have been updated by following the process described in https://github.com/bonitasoft/bonita-documentation-site/blob/master/docs/documentation-components-and-versions.adoc#mark-version-as-out-of-support.

We don't accept contribution for this version. So remove workflows and the PR template.


tbouffard:
Notice that the following is not fully documented in https://github.com/bonitasoft/bonita-documentation-site/blob/master/docs/documentation-components-and-versions.adoc#mark-version-as-out-of-support
This has been done like this for the 7.9 version (7788fcdbc1faff52aa7973a0514c0b707399b963), so I replicate the changes.
I will create a doc PR in the doc-site repo soon.

alachambre (approved):

PR #1926, opened 2022-02-07
chore(release): add missing runtime bug in release note 2021.2-u3

baptistemesta:
Relates to [RUNTIME-302](https://bonitasoft.atlassian.net/browse/RUNTIME-302)


baptistemesta:
@danila-m this was missing from the release note. The fix is included in that version

alachambre (approved):

PR #1927, opened 2022-02-08
Fix(timerEventTrigger): Add filter params for the GET method. 

cleosurville:
Add documentation for the filters allowed in the api GET method for timerEventTrigger.

Closes [RUNTIME-838](https://bonitasoft.atlassian.net/browse/RUNTIME-838)

### DIFF
--- a/modules/api/pages/bpm-api.adoc
+++ b/modules/api/pages/bpm-api.adoc
@@ -5250,6 +5250,10 @@ _Example_: `/API/bpm/timerEventTrigger?caseId=4025&p=0&c=10&`
 `GET`
 * *Data Params*
  ** `caseId`: ID of the case (Process instance)
+ ** f: Filters allowed in the api url.
+  *** name: The name of BPM timer event trigger. ex: f=name=MyBoundaryEvent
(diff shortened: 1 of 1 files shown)


modules/api/pages/bpm-api.adoc
```diff
@@ -5250,6 +5250,10 @@ _Example_: `/API/bpm/timerEventTrigger?caseId=4025&p=0&c=10&`
 `GET`
 * *Data Params*
  ** `caseId`: ID of the case (Process instance)
+ ** f: Filters allowed in the api url.
+  *** name: The name of BPM timer event trigger. ex: f=name=MyBoundaryEvent
+  *** executionDate: Timestamp (in millisecond) of the date of the BPM timer event trigger.  ex: f=executionDate=1644331888385
```
educhastenier, line 5255: Is it really useful to search for timers that would trigger at a specific timestamp that we already know about?

cleosurville: In fact i don't know, i just documented the current possibilities. But why would it be possible to do it in the engine otherwise ?

PR #1928, opened 2022-02-09
chore(dependencies): adding portal-js dependencies report for version…

baptistemesta:
Manually created, 
The job did not do it


baptistemesta:
The changes are really significant!

PR #1939, opened 2022-02-16
feat(bdm): Document of eager & lazy loading works

danila-m:
Add a technical section about how eager & lazy loading works "under the hood".

Relates to [RUNTIME-801](https://bonitasoft.atlassian.net/browse/RUNTIME-801)

### DIFF
--- a/modules/data/pages/define-and-deploy-the-bdm.adoc
+++ b/modules/data/pages/define-and-deploy-the-bdm.adoc
@@ -152,7 +152,7 @@ When you create a process that uses a business object with a composition or aggr
 
 [#lazy_eager_loading]
 
-=== Loading in Hibernate
+=== Loading BDM objects
 
 There are two design patterns for loading business objects: *LAZY* or *EAGER*. +
 
@@ -179,8 +179,8 @@ The tradeoff between those two design patterns is execution time vs memory consu
 
 [WARNING]
 ====
-If you are not sure about the final size of BDM schema, we do *strongly* recommend to use the default *LAZY* loading design pattern. Indeed, *EAGER* loading design pattern can have a huge impact on the Bonita Runtime performance, due to the additional data load and information maintenance in memory. +
-Based on your tecnical analysis, if you still decide to use *EAGER* design pattern, *we strongly recommend thorough testing with a test dataset having the size of your production's expected dataset*.
+If you are not sure about the final size of BDM schema, we do *strongly* recommend to use the default *LAZY* loading design pattern. Indeed, *EAGER* loading design pattern can have a huge impacts on the Bonita Runtime performance, due to the additional data load and information maintenance in memory. +
+If you still decide to use *EAGER* design pattern, *we strongly recommend thorough testing with a test dataset having the size of your production's expected dataset*.
 ====
 
 === Multi-instantiation using business object variables in a list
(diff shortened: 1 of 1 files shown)


alachambre:
Why do you put such a large content in a note? 
Admonitions are meant to be used for small et concise content, a sentence that should not be missed. 
A complete paragraph should be written in a dedicated section.

danila-m:
> Admonitions are meant to be used for small et concise content, a sentence that should not be missed.
> A complete paragraph should be written in a dedicated section.

The point was to put a "technical/under the hood" section, rather than simply another text paragraph. It's not exactly required knowledge to use the feature, but nice to know & can be very useful, for support & future devs for example.

Maybe another flag than note? I don't know what's available.

modules/data/pages/define-and-deploy-the-bdm.adoc
```diff
@@ -168,6 +168,17 @@ There are two options for loading complex business objects: *lazy* or *eager*. +
 
 Choosing the right option is not easy. The default option is lazy loading, but you may override this for any business object by configuring it to use eager loading when you specify the object relationship properties. +
 
+[Note]
```
alachambre:
```suggestion
[NOTE]
```

alachambre:
> > Admonitions are meant to be used for small et concise content, a sentence that should not be missed.
> > A complete paragraph should be written in a dedicated section.
> 
> The point was to put a "technical/under the hood" section, rather than simply another text paragraph. It's not exactly required knowledge to use the feature, but nice to know & can be very useful, for support & future devs for example.
> 
> Maybe another flag than note? I don't know what's available.

You can refer to the official documentation to know the available admonitions: https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/#admonition-types

My point is that admonitions are not suitable for large paragraph, it is not confortable to read a lot of technical stuff in a coloured square. It is made to catch your eyes and make you read a short and important point.
In the Bonita documentation we use to many admonitions for nothing, it make the whole documentation unpleasant to read. 
So I suggest to keep it for when it seems necessary and put large paragraph in simple sections.  
But if you think that an admonition is fine here keep it, that's just a suggestion. 

Here is the preview if you missed it, in the page there is two large admonitions one after the other.. 
Capture d’écran 2022-02-22 à 15 09 52

Ps: You must use caps for admonitions, I fixed it for you

alachambre:
As a terrible example of a page with too many admonitions, check this one for ex, I think you will get my point: 

https://documentation.bonitasoft.com/bonita/2021.2/getting-started/draw-bpmn-diagram

akantcheff (approved):

vbotnari12 (approved):
all good for me!

modules/data/pages/define-and-deploy-the-bdm.adoc
```diff
@@ -152,26 +152,35 @@ When you create a process that uses a business object with a composition or aggr
 
 [#lazy_eager_loading]
 
-=== Loading
+=== Loading BDM objects
 
-There are two options for loading complex business objects: *lazy* or *eager*. +
+There are two design patterns for loading business objects: *LAZY* or *EAGER*. +
```
educhastenier, line 157: You should precise that it only concerns aggregation, if my memory is good

vbotnari12: what exactly? the design patterns or loading business objects?

alachambre: Your memory is NOT good, a complex field used in composition or in aggregation can be loaded lazily or eagerly. 

But I do not recommend to use the word _design pattern_ here. In computer science, a design pattern is a reusable solution to a commonly occurring problem.
The way to load a complex field has nothing to do with a design pattern IMHO, using this word will lead to confusion.
And you do not really load business objects, you load complex fields of business objects. 

You should update this sentence (in an other pr since this once has been merged), I believe that being accurate is the best way to help users understand complex concepts.

PR #1940, opened 2022-02-17
Update(update-studio): internet mandatory caution

marielle-spiteri:
* because it is


alachambre:
@florentsarat could you please check this pull request?

PR #1942, opened 2022-02-18
feat(test): Add reference to the Bonita Test Toolkit

florentsarat:


modules/process/pages/process-testing-index.adoc
```diff
@@ -18,6 +18,23 @@ If you have already created your own xref:ROOT:actors-index.adoc[actors] for you
 . Select the process pool
 . Click on the *Configure* icon of the Studio coolbar and configure the _Actor mapping_
 
+== Integration Testing
+
+[NOTE]
+====
+
```
alachambre:
```suggestion
```

modules/process/pages/process-testing-index.adoc
```diff
@@ -18,6 +18,23 @@ If you have already created your own xref:ROOT:actors-index.adoc[actors] for you
...
+For Enterprise, Performance, Efficiency and Teamwork editions only.
+====
+
+To write automated integration test scenarios for your automation project you can use the Bonita Test Toolkit. It provides powerful methods to execute all the steps of a process and check the corresponding outputs (statuses, business data, tasks used)
+
+It can be used in any IDE and can easily be integrated into a Continuous Integration.
```
alachambre: DO not use easily in a technical documentation, you do not know the level of your reader. 
Especially when it's absolutely not easy, even for us ^^

```suggestion
It can be used in any IDE and can be integrated into a Continuous Integration.
```

alachambre (changes requested):

modules/process/pages/process-testing-index.adoc
```diff
@@ -18,6 +18,23 @@ If you have already created your own xref:ROOT:actors-index.adoc[actors] for you
...
+
+More information in the dedicated documentation: https://documentation.bonitasoft.com/test-toolkit/1.0/process-testing-overview[Bonita Test Toolkit].
```
alachambre: You are not allowed to use the URL of the docsite like this. And you should only reference the version of a component in some specific situations, but not here.
The test toolkit is a composant of the doc sources, reference it using the antora macro as any other page. 
More details here: https://docs.antora.org/antora/latest/page/page-links/#component-versions

```suggestion
More information in the dedicated documentation: xref:test-toolkit:ROOT:process-testing-overview.adoc[Bonita Test Toolkit] 
```

alachambre (approved):

PR #1943, opened 2022-02-22
Fix(troubleshooting sections 2nd batch): add icons and standardized sections

NathalieC:
* in Bonita Studio pref and config: add image for troubleshooting
* in Bonita Studio pref and config: add line breaks in troubleshooting section

### DIFF
--- a/modules/identity/pages/single-sign-on-with-oidc.adoc
+++ b/modules/identity/pages/single-sign-on-with-oidc.adoc
@@ -315,7 +315,7 @@ com.bonitasoft.level = FINE
 *Symptom:* +
 After configuring OIDC SSO in Bonita, the Bonita login page appears and you are not redirected to the SSO login page. +
 
-*Possible Solutions:*
+*Possible solutions:*
 
 * Check all the Bonita configuration settings are correct.
 * Make sure `setup[.sh][.bat] push` has been executed and the server restarted after the changes.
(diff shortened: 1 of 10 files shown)


modules/identity/pages/single-sign-on-with-oidc.adoc
```diff
@@ -286,7 +286,9 @@ To setup Bonita for OIDC logout:
 If the logout flow supported by your OIDC provider is not the same as the one supported by Bonita platform, the preferred solution to handle it anyway is to intercept the requests to /logoutService and handle the logout programmatically.
 ====
 
-== Troubleshoot
```
alachambre, line 289: Why is this section not at the bottom of the page, like in all other pages?

modules/identity/pages/single-sign-on-with-saml.adoc
```diff
@@ -339,7 +339,9 @@ To setup Bonita for global logout:
 If the single logout flow supported by your IdP is not the same as the one supported by Bonita platform, the preferred solution to handle it anyway is to intercept the requests to /logoutService and handle the logout programmatically.
 ====
 
-== image:images/troubleshooting.png[troubleshooting-icon] Troubleshooting
+== image:ROOT:images/troubleshooting.png[troubleshooting-icon] Troubleshooting
```
alachambre, line 379: Why is this section not at the bottom of the page?

alachambre (changes requested):
I think the two sso troobleshooting sections should be moved at the bottom of the pages.

modules/identity/pages/single-sign-on-with-oidc.adoc
NathalieC: thanks. done.

modules/identity/pages/single-sign-on-with-saml.adoc
NathalieC: thx. done

PR #1976, opened 2022-03-08
doc(MigrationTool) : Removing upgrade from the documentation

vbotnari12:
Removing "upgrade" from the documentation of Migration Tool, because upgrade it's subscription change, not version change.

### DIFF
--- a/modules/ROOT/pages/migrate-from-an-earlier-version-of-bonita-bpm.adoc
+++ b/modules/ROOT/pages/migrate-from-an-earlier-version-of-bonita-bpm.adoc
@@ -36,10 +36,10 @@ You are recommended not to start 7.0.0 after you migrate to it, but to proceed i
 ====
 
 *RDBMS requirements:*
-The version targeted may not support the version of the database that is being migrated. You may then need to update the version of your database prior to running the migration tool.
+The version targeted may not support the version of the database that is being migrated. You may then need to upgrade the version of your database prior to running the migration tool.
 
 * Please check the xref:hardware-and-software-requirements.adoc[database requirements].
-* If you need to update your database:
+* If you need to upgrade your database:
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/migrate-from-an-earlier-version-of-bonita-bpm.adoc
```diff
@@ -36,18 +36,18 @@ You are recommended not to start 7.0.0 after you migrate to it, but to proceed i
 ====
 
 *RDBMS requirements:*
-The version targeted may not support the version of the database that is being migrated. You may then need to upgrade the version of your database prior to running the migration tool.
+The version targeted may not support the version of the database that is being migrated. You may then need to update the version of your database prior to running the migration tool.
```
educhastenier: @vbotnari12 We should check if database vendors use the term update of upgrade of migrate for the Database server "mise à jour"...

modules/ROOT/pages/migrate-from-an-earlier-version-of-bonita-bpm.adoc
```diff
@@ -36,18 +36,18 @@ You are recommended not to start 7.0.0 after you migrate to it, but to proceed i
...
 
 * Please check the xref:hardware-and-software-requirements.adoc[database requirements].
-* If you need to upgrade your database:
+* If you need to update your database:
```
educhastenier: same remark

educhastenier:
@vbotnari12 ❓ Is it normal you make this PR on a 2021.1 base?

modules/ROOT/pages/migrate-from-an-earlier-version-of-bonita-bpm.adoc
vbotnari12: nice catch. AWS uses upgrade, Oracle uses both - upgrade and migration.

PR #1986, opened 2022-03-10
feat(dynamic permissions): improve release note

abirembaut:
* we are satisfied with the performances so I don't think we should ask to bench
* the decision to activate the dynamic checks is more related to applications compatibility I think


modules/ROOT/pages/release-notes.adoc
```diff
@@ -56,7 +56,7 @@ These credentials can be consulted and modified:
 === Dynamic authorization checking active by default (Subscription editions)
 
 xref:identity:rest-api-authorization.adoc#dynamic_authorization[Dynamic authorization checking] is now only available in the *Enterprise* editions, and is also *active by default* for new installations. +
-Existing installations that update to 2022.1 stay *unchanged*, but we encourage customers to activate dynamic authorization checking, once they have benched that the performance cost is reasonable. +
+Existing installations that update to 2022.1 stay *unchanged*, but we encourage customers to xref:identity:rest-api-authorization.adoc#dynamic_authorization[activate dynamic authorization checking], once they have validated that the default dynamic rules provided in Bonita Runtime are complient with their applications (if not, rules and/or application pages need to be adapted). +
```
danila-m:
```suggestion
Existing installations that update to 2022.1 will stay unchanged (meaning dynamic permissions will not be automatically activated or deactivated upon update), but we strongly encourage customers to xref:identity:rest-api-authorization.adoc#dynamic_authorization[activate dynamic authorization checking], once they have validated that the default dynamic rules provided in Bonita Runtime are compliant with their applications. If not, rules and/or application pages will need to be adapted. +
```

danila-m (changes requested):
Better phrasing

PR #1988, opened 2022-03-11
fix(docker): replace varversion w bonitaversion

marielle-spiteri:
* replace varversion w bonitaversion in Subscription release docker command line


educhastenier (approved):
Looks good to me :+1:

PR #2009, opened 2022-03-22
(fix): remove extra ";"

laurentleseigneur:
following @tbouffard advise

### DIFF
--- a/modules/ROOT/pages/event-handlers.adoc
+++ b/modules/ROOT/pages/event-handlers.adoc
@@ -65,7 +65,7 @@ pom.xml :
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


tbouffard (changes requested):
The issue exist in other pages and starting from Bonita 7.11, so the fix should target this version.
The proposed fixed changes the rendered schemaLocation: we should see a space

educhastenier:
I do not see any `;` change in this PR 🤔

tbouffard:
@educhastenier the internal reference for the issue is [RUNTIME-969](https://bonitasoft.atlassian.net/browse/RUNTIME-969). It provides more details about the problem and the result of a quick investigation.

laurentleseigneur:
rebased PR on 7.11 branch as requested by @tbouffard

tbouffard:
The rebase on 7.11 is ok, but the fix is not accurate. 
There are also other pages impacted by the issue, see [RUNTIME-969](https://bonitasoft.atlassian.net/browse/RUNTIME-969)

laurentleseigneur:
seems i didn't get the way to fix it by removing the `` in XSD section ?

PR #2010, opened 2022-03-22
doc(log4j): improve doc in docker volume mount explanations

educhastenier:


modules/runtime/pages/bonita-docker-installation.adoc
```diff
@@ -379,9 +379,9 @@ Since Bonita 7.9 BONITA_SERVER_LOGGING_FILE and BONITA_SETUP_LOGGING_FILE can be
 The logger can be configured by mounting a volume on folder `/opt/bonita/conf/logs` containing the configuration files.
 
 the volume must contain the 2 files
-https://github.com/bonitasoft/bonita-distrib/blob/dev/tomcat-resources/tomcat-distrib-for-bonita/src/main/resources/tomcat/server/conf/log4j2-loggers.xml[log4j2-loggers.xml]
+https://raw.githubusercontent.com/bonitasoft/bonita-distrib/dev/tomcat-resources/tomcat-distrib-for-bonita/src/main/resources/tomcat/server/conf/log4j2-loggers.xml[log4j2-loggers.xml]
 and
-https://github.com/bonitasoft/bonita-distrib/blob/dev/tomcat-resources/tomcat-distrib-for-bonita/src/main/resources/tomcat/server/conf/log4j2-appenders.xml[log4j2-appenders.xml]
+https://raw.githubusercontent.com/bonitasoft/bonita-distrib/dev/docker/files/log4j2/log4j2-appenders.xml[log4j2-appenders.xml]
```
tbouffard: ℹ️ I suggest to use a tag instead of the dev branch to stick to the 2022.1 version. We have a AsciiDoc attribute which can be used for this purpose.

educhastenier: What attribute are you talking about? bonitaVersion? bonitaTechnicalVersion? 🤔 ...

To the contrary, I think pointing to the master or dev branch keeps the moving reference up to date with the doc.
Maybe **master** would better fit...

tbouffard: Today the 2022.1 documentation relates to the code in the master/dev (whatever) branch, but this won't be the case in 6 months because the branch will match the next version. So any file move, change, rename, removal, .... will break the documentation.
So from a maintenance point of view, this is an issue. 
In addition, the link is targeting the community repository: there is generally no change in the master branch (except major fixes) so the file won't be updated for a given version.

About the AsciiDoc attribute, I had in mind those available in the `antora.yml` file, whatever it is as long it is accurate for the link here.

educhastenier: Ok, I understand what you say.
Problem with a tag reference, is that on later versions, the link should be updated with new tag reference, if the content of the file log4j2-appenders.xml or log4j2-loggers.xml change.
And most probably, we will forget to update it.
With a reference to **dev** however, it will always be up-to-date. Problem could then be with older versions of the doc, pointing to dev branch, but for those files, I believe there may not be major incompatibilities with newer versions of the content.
So I suggest to use **dev** reference.

Do you agree or do you still want to use a tag?

tbouffard: The `bonitaTechnicalVersion` Asciidoc attribute is supposed to be set to the latest tag of a given version available for Bonita community, so creating an url using this attribute will always be accurate as far of the attribute is updated when new Community maintenance version is available (this almost never happens).
This attribute is set according to the branch of the documentation, so it will be accurate in 2022.2 as well.
What I suggest is to use this attribute in the url: at site generation, it will be substituted. We already do that for javadoc links and we haven't seen any issue nor out of date refs: `https://javadoc.bonitasoft.com/api/{javadocVersion}/org/bonitasoft/engine/filter/UserFilter.html`

I don't understand why you still want to push for the branch solution.
The content of master and dev branch is always evolving as I previously stated. In 1 year, if the log configuration changes (and it will - think about moving to spring-boot, there won't be any log4j2 configuration files), the files won't be available and the old documentation will be broken. This won't happen with links using git tags (created with AsciiDoc attribute substitution).

I let your team decide what is the best.

educhastenier: Sorry, you're perfectly right: the `bonitaTechnicalVersion` is different for each version, so there is no problem using / maintaining it.
Thx. I'll do that.

tbouffard: No problem, next time I will be more explicit in my 1st comment and this will avoid comments ping-pong 😸

PR #2011, opened 2022-03-22
fix(procdata): do not store BO in proc variable

marielle-spiteri:
* users must be warned that it's not a supported, and it's a bad practice. It's not obvious enough in the doc.

### DIFF
--- a/modules/ROOT/pages/specify-data-in-a-process-definition.adoc
+++ b/modules/ROOT/pages/specify-data-in-a-process-definition.adoc
@@ -24,6 +24,11 @@ Best practice:
 * Define a variable at the lowest level possible, using a process variable only when required for the process flow.
 * Use the minimum number of variables possible; having an unnecessarily large set of variables has an impact on performance.
 
+[WARNING]
+====
+Business Object should not be stored in Process or activity data. They can only be stored in Business data and *Form* transient data.
+====
+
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/specify-data-in-a-process-definition.adoc
```diff
@@ -24,6 +24,11 @@ Best practice:
 * Define a variable at the lowest level possible, using a process variable only when required for the process flow.
 * Use the minimum number of variables possible; having an unnecessarily large set of variables has an impact on performance.
 
+[WARNING]
+====
+Business Object should not be stored in Process or activity data. They can only be stored in Business data and *Form* transient data.
```
rbioteau: What are Form transient data ? Did you mean transient data ?

marielle-spiteri: Form variable are always transient so I guess I don't need that. 
BO should never be stored in activity's transient variable. 

Maybe this will be clearer, what do you think?: 
Business Object should not be stored in Process or activity variable. They can only be stored in Business data variable and form variable.

rbioteau: Yeah, it is a little confusing. I would stick with Business Object = Business variables.

marielle-spiteri: ok, I made a new commit with clearer sentence.

vbotnari12 (approved):
all good +1

PR #2012, opened 2022-03-23
fix(code sample): Remove extra `;`in XML file

laurentleseigneur:
apply same fix that for #2009


tbouffard (changes requested):
I don't understand the purpose of this PR knowing #2009 is already intended to manage the fix in 7.11.

laurentleseigneur:
it's not the same file

tbouffard:
> it's not the same file

fine, I didn't get it. 😆 
I suggest to do the fixes of all doc pages in #2009 as the current proposed changes are not working.

PR #2017, opened 2022-03-29
fix(OIDC): fix group path

laurentleseigneur:
example with automatic user création doesn't work with default ACME organization
path is lower case


abirembaut:
LGTM

PR #2018, opened 2022-03-29
chore(OIDC): doc for reverse proxy configuration

laurentleseigneur:
When reverse proxy is responsible of the HTTPS part and Tomcat remains in HTTP, a specific config should be added to tomcat to use the correct procolHeader forwarded by the proxy

* add support for ...
* fix ... behaviour
* refactor ....


abirembaut:
@laurentleseigneur 
reworded here : #2019 as there was already a warning block explaining half of the solution

PR #2020, opened 2022-03-30
feat(RN): remove documentation about unsupported sysprop

baptistemesta:
That system property allows to log directly in console output. It's used by studio but it's not easier than using the default way to configure it: throw log4j2-appenders


passga (approved):
LGTM!!!

PR #2029, opened 2022-04-05
doc: Add missing title

benjaminParisel:
* Without a title in top of the file, the tab title in your browser display Untitled

Before:

image

After:

image


modules/ROOT/pages/activities-tasks-flow-nodes-api.adoc
```diff
@@ -1,3 +1,4 @@
+= Activities, Tasks and Flow nodes
```
NathalieC:
```suggestion
= Activities, Tasks, and Flow nodes
```

NathalieC: add a coma after before "and"

PR #2030, opened 2022-04-07
Update getting-started

NathalieC:
* fix Bonita Studio Installation description
* add line break and fix do not worry
* fix bold / italic / quote marks for items to click


NathalieC:
Will retry later on.

PR #2034, opened 2022-04-11
update(product versioning): element and table

NathalieC:
* change table data from 2021.2 to 2022.1

### DIFF
--- a/modules/version-update/pages/product-versioning.adoc
+++ b/modules/version-update/pages/product-versioning.adoc
@@ -36,37 +36,17 @@ Here is a Bonita Platform version / Technical Id / UI Designer version correspon
 |===
 | Release date | Bonita Platform version | Technical Id | UI Designer version
 
-| 2022-03-03
-| 2021.2-u4
-| 7.13.4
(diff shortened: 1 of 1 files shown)


tbouffard:
@NathalieC the build of the site preview is in error because a page in the 2021.2 branch (so the branch of the later versions as well) contains a duplicated block id. This will be fixed by #2033

educhastenier:
> @NathalieC the build of the site preview is in error because a page in the 2021.2 branch (so the branch of the later versions as well) contains a duplicated block id. This will be fixed by #2033

✅ Fixed and rebased

tbouffard (commented):
✔️ the table data renders correctly

image

tbouffard (changes requested):
@NathalieC mentioned me privately that the `artifact` to `element` renaming should also applied to `2021.2`.
So this PR must be amended to remove the renaming here, and a new PR targeting `2021.2` with the related changes must be created (see #2307)

tbouffard (approved):
LGTM, the renaming has been removed

PR #2040, opened 2022-04-13
feat(docker): Update docker install

danila-m:
* correct postgres startup command
* homogenize the license folder to ~/bonita-lic
* remove the docker-compose: it does not work in 2022.1 anymore, there is no easy way to make it work - so we remove it, for lack of a better solution.

Closes [RUNTIME-889](https://bonitasoft.atlassian.net/browse/RUNTIME-889)

### DIFF
--- a/modules/runtime/pages/bonita-docker-installation.adoc
+++ b/modules/runtime/pages/bonita-docker-installation.adoc
@@ -169,8 +169,10 @@ services:
       - -c
       - max_prepared_transactions=100
   bonita:
-    image: bonita:7.14.0
-    hostname: custom-hostname.example.com
+    image: quay.io/bonitasoft/bonita-subscription:pass:a[{bonitaVersion}]
+    hostname: <hostname>
+    volumes:
+      - ~/bonita-lic:/opt/bonita_lic/
     ports:
       - 8080:8080
     environment:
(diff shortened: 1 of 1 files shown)


modules/runtime/pages/bonita-docker-installation.adoc
```diff
@@ -145,75 +145,11 @@ You can run the image with the following command:
 
 [source,bash]
 ----
-docker run --name mydbpostgres -h <hostname> -v <your-bonita-license-folder>:/opt/bonita_lic/ -d bonitasoft/bonita-postgres:12.6
+docker run --name mydbpostgres -d bonitasoft/bonita-postgres:12.6
 ----
 
 This image is built from the following https://github.com/Bonitasoft-Community/bonita-database-docker/tree/main/postgres/12[GitHub repository], which can be further adapted/customized to suit your needs.
 
-
```
educhastenier, line 153: With Baptiste, we found a way to fix docker-compose, so not need to remove this.
See https://hub.docker.com/_/bonita
Moreover, the fix for the docker-compose should be done on 2021.1 or 2021.2 (I don't know exactly, the version where we removed psql client from Bonita docker image)

danila-m: I changed the docker compose to the new one.
The postgres client was removed only in 7.14, so 2022.1 is the correct branch

PR #2042, opened 2022-04-19
fix(cust-auth): use java 11 and specific bonita version in example

marielle-spiteri:
* Fix:  <bonita-server.version>pass:a[{bonitaTechnicalVersion}</bonita-server.version>


rbioteau (commented):
I'm not sure if it was a typo, but more a version range.

tbouffard (approved):

PR #2057, opened 2022-05-12
feat(http): clarify HTTP Basic Auth configuration

educhastenier:


vbotnari12 (approved):
Ok, after removing passwords from the release note.

educhastenier:
> Ok, after removing passwords from the release note.

@vbotnari12 Do you mean you are ok to keep default password values everywhere BUT in the release notes? Why?

vbotnari12:
> > Ok, after removing passwords from the release note.
> 
> @vbotnari12 Do you mean you are ok to keep default password values everywhere BUT in the release notes? Why?
For the very simple reason that in the release note we put general information and if needed, direct users for details in the documentation. Ideally, there should not be passwords in the documentation (nor install/install), but that's a topic we are not going to address here.

ugaston (approved):
This is really helpful!!
It looks good for me.

Thanks Manu!

PR #2072, opened 2022-06-20
doc(deps): indicate that new Maven Repo is necessary

educhastenier:
if pulling bonita-server(-sp).jar

Relates to [RUNTIME-991](https://bonitasoft.atlassian.net/browse/RUNTIME-991)

### DIFF
--- a/modules/ROOT/pages/release-notes.adoc
+++ b/modules/ROOT/pages/release-notes.adoc
@@ -171,8 +171,42 @@ Bonita Runtime produces a warning log if the filter is set and ignores it.
 
 The tenant creation method has been deprecated following our 2021.1 decision to deprecate the Multi-Tenants architecture.
 
+=== Potential need of declaring external Maven Repository
+
+From version 2022.1-u2, if you embed *bonita-server(-sp).jar* in your application, you need to declare some new Maven Repository in you pom.xml / build.gradle. This is due to some Security Fix that requires a dependency that is NOT present in default Maven Repository (aka Maven Central). To do so, add the following code:
+
+==== Pom.xml
+
+```xml
+<repositories>
+    [...]
+    <repository>
+        <id>terracotta.org</id>
+        <name>terracotta.org</name>
+        <url>https://repo.terracotta.org/maven2/</url>
+    </repository>
+</repositories>
+```
+
+==== build.gradle
+
+[source,groovy]
+----
+repositories {
+    mavenCentral()
+    maven {
+        url 'https://repo.terracotta.org/maven2/'
+    }
+}
+----
 
 == Bug fixes
+
+=== Fixes in Bonita 2022.1-u2 (2022-0X-YY)
+
+==== Fixes in Bonita Runtime (including Bonita Applications)
+
+* RUNTIME-991  - [Security] Update several dependencies for 2022.1-u2 (see xref:#_potential_need_of_declaring_external_maven_repository[specific note]])
 
 === Fixes in Bonita 2022.1-u1 (2022-05-03)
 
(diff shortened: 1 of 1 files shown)


vbotnari12 (approved):
OK for me.

modules/ROOT/pages/release-notes.adoc
```diff
@@ -171,9 +171,43 @@ Bonita Runtime produces a warning log if the filter is set and ignores it.
 
 The tenant creation method has been deprecated following our 2021.1 decision to deprecate the Multi-Tenants architecture.
 
+=== Potential need of declaring external Maven Repository
+
+From version 2022.1-u2, if you embed *bonita-server(-sp).jar* in your application, you need to declare additional Maven Repository in your pom.xml / build.gradle. This is due to Security Fixes that required the use of a dependency NOT present in default Maven Repository (aka Maven Central). To do so, add the following code:
```
tbouffard: 🔥 Adding repositories in pom.xml has been described as a bad practise for years
- https://blog.sonatype.com/2009/02/why-putting-repositories-in-your-poms-is-a-bad-idea/ (2009)
- https://maven.apache.org/repository-management.html

The remark here should at least advise usage of repository manager. This is not the only place where this should advice!

tbouffard (changes requested):

tbouffard:
ℹ️ the build of the preview is failing due to the configuration of the GitHub workflow.
For xref validation between component, we add the 2022.1 branch of bonita because the cloud component references it directly.
We don't have logic in the workflow to not add the 2022.1 branch when the PR branch is based on this branch. So the content of the 2022.1 and PR branches are almost duplicated leading to an Antora error
> FATAL (antora): Duplicate nav in 2022.1@bonita: modules/ROOT/build-a-process-for-deployment-nav.adoc
  1: modules/ROOT/build-a-process-for-deployment-nav.adoc in https://github.com/bonitasoft/bonita-doc.git (ref: 2022.1)
  2: modules/ROOT/build-a-process-for-deployment-nav.adoc in https://github.com/bonitasoft/bonita-doc.git (ref: fix/indicate_ehcache_need_for_new_repo)

We are working on this topic as part of https://github.com/bonitasoft/bonita-documentation-site/issues/326

modules/ROOT/pages/release-notes.adoc
```diff
@@ -171,9 +171,43 @@ Bonita Runtime produces a warning log if the filter is set and ignores it.
...
+
+[source,groovy]
+----
+repositories {
+    mavenCentral()
+    maven {
+        url 'https://repo.terracotta.org/maven2/'
+    }
+}
+----
```
educhastenier, line 205:
```suggestion
----

Of course, you can abstract this additional complexity by putting this configuration in your https://maven.apache.org/repository-management.html[enterprise repository manager] which acts a proxy to any external repositories you may have.
```

modules/ROOT/pages/release-notes.adoc
```diff
@@ -171,9 +171,45 @@ Bonita Runtime produces a warning log if the filter is set and ignores it.
...
+
+Of course, you can abstract this additional complexity by putting this configuration in your https://maven.apache.org/repository-management.html[enterprise repository manager] which acts a proxy to any external repositories you may have.
```
tbouffard: This applies to both maven and gradle so putting it before the examples will make the message more visible. Currently, it seems to me that it only applies to gradle.

ℹ️ The repository manager usage is not only for simplification, but if the Runtime Team agrees with the current wording, it's fine to me.

tbouffard (approved):

modules/ROOT/pages/release-notes.adoc
educhastenier: Ok, I'll take into consideration this first remark (and let the second one like that 😉 )

educhastenier: done

tbouffard: thanks

PR #2075, opened 2022-06-24
chore: remove cloud branch reference

benjaminParisel:
* remove cloud branch reference to avoid duplicate version on 2022.1 branches on pr-preview build


tbouffard (approved):

tbouffard:
⚠️ let's merge this PR then propagate and fix in 2022.2 that will require we restore the 2022.1 branch

PR #2085, opened 2022-07-18
update(UpdateTool): Add manual operations after using Update Tool

4rthurLegrand:
* Adding details about manual operations to do after using Update Tool that are available on the Migration Tool page, but not Update Tool page.


vbotnari12 (approved):
All good for merging.

PR #2105, opened 2022-08-11
doc(docker-image): versions available on quay

vbotnari12:
Precise the versions that are available on quay.io so that customers will not search for a beta version.


tbouffard:
The "PR preview" build failure is not due to the changes proposed in this PR.
Test-toolkit@1.0 now requires bonita@2022.2 for xref validation and this version is not available when currently building preview for bonita@2021.2.
This should be fixed by #2107

PR #2125, opened 2022-08-31
fix(docker-compose): escape  $ to have a great display

passga:
escape  $ to fix docker-compose to  have a great display in the doc site


educhastenier (approved):
Rendering looks good on the preview 👍🏽

PR #2129, opened 2022-09-06
docs(2022.2): update product versioning information

benjaminParisel:
* Only merged this PR when GA is released


vbotnari12 (approved):
OK for GA

PR #2166, opened 2022-09-23
docs: improve PR title check

benjaminParisel:
* Use more common action to detect if the title is following the convention

### DIFF
--- a/.github/workflows/contribution-check.yml
+++ b/.github/workflows/contribution-check.yml
@@ -5,14 +5,13 @@ on:
     types: [opened, edited, synchronize]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/contribution-check.yml
```diff
@@ -5,14 +5,13 @@ on:
     types: [opened, edited, synchronize]
 
 jobs:
-  commit-message-check:
+  lint-pr:
```
tbouffard: I suggest something which more explicit.
```suggestion
  pr-title:
```

tbouffard (approved):
LGTM

PR #2167, opened 2022-09-23
chore(png): Compressed png files

jeromecambon:
Compressed all the png images, using the optipng CLI tool.
We are saving more than 3Mb (12Mb -> 8.7Mb)

Covers https://github.com/bonitasoft/bonita-documentation-site/issues/220


tbouffard (changes requested):
The conversion looks good.
However, I currently don't understand the whole conversion strategy across branches, so I suggest we don't merge this PR for now to avoid future merge issue: https://github.com/bonitasoft/bonita-documentation-site/issues/220#issuecomment-1263664992

jeromecambon:
> The conversion looks good.
> However, I currently don't understand the whole conversion strategy across branches, so I suggest we don't merge this PR for now to avoid future merge issue: [bonitasoft/bonita-documentation-site#220 (comment)](https://github.com/bonitasoft/bonita-documentation-site/issues/220#issuecomment-1263664992)

ok, we'll discuss about this. As I commented in the issue, not sure it deserve to be applied in older branches.

tbouffard (approved):
I still disagree with the strategy that has been chosen: IMHO, the oldest branch should have been updated first.
But like I said in https://github.com/bonitasoft/bonita-documentation-site/issues/220#issuecomment-1266886553, I don't veto this PR as it contains some valuable work.
I let someone else review this PR and decide if we merge it or not.

PR #2168, opened 2022-09-26
docs: Update widgets page

ghost:
Remove spurious old fa-exclamation-triangle symbol that seems not interpreted by antora (and covered by the above warning anyway)


vbotnari12:
Conflict should be removed.

PR #2172, opened 2022-09-28
fix(FileViewer widget): Add missing limitation for the URL property

jeromecambon:
Covers [UID-703](https://bonitasoft.atlassian.net/browse/UID-703)

### DIFF
--- a/modules/ROOT/pages/widgets.adoc
+++ b/modules/ROOT/pages/widgets.adoc
@@ -441,7 +441,7 @@ A options toolbar exposed to end users can be customized and localized with the 
 [.troubleshooting-section]
 --
 [.symptom]
-FileViewer widget does not preview a file (although a point on a supported format)
+File Viewer widget does not preview a file (although I point on a supported format)
 
 [.symptom-description]
 I set the URL property to a pointer on a pdf or image file, but I can't see the preview: I get a "Preview is not available" message.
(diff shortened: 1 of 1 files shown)


benjaminParisel:
:question: @jeromecambon why you didn't use the troubleshooting syntax to explain how the user could fall into this problem and a way to fix it?

vbotnari12 (approved):
ok for me

jeromecambon:
> ❓ @jeromecambon why you didn't use the troubleshooting syntax to explain how the user could fall into this problem and a way to fix it?

Using the troubleshooting syntax in the File Viewer section would be too proeminent IMO.
So I initiated an "Troubleshooting UI Designer widgets" section at the end of the page, which contains for now this specific case. The idea is too add new items here when needed.

benjaminParisel:
1. I think the warning section is totally a section for troubleshooting. 
2. A whitespace is missing between "limitations" and "the bullets list". On the preview, a '-' is displaying instead of a bullet.

jeromecambon:
> * I think the warning section is totally a section for troubleshooting.
> * A whitespace is missing between "limitations" and "the bullets list". On the preview, a '-' is displaying instead of a bullet.

IMO it's better to centralize the troubleshooting stuff in the same place, to avoid to "pollute" the widget sections.
I don't see a whitespace missing... 
The usual bullet syntax ('*') was not working in the Warning section, that's why I put dashes.

benjaminParisel:
> The usual bullet syntax ('*') was not working in the Warning section, that's why I put dashes.
You need to use the 
`WARNING
===
While werewolves are hardy community members, keep in mind the following dietary concerns:

* They are allergic to cinnamon.
* More than two glasses of orange juice in 24 hours makes them howl in harmony with alarms and sirens.
* Celery makes them sad.
===
`
image

julienmege:
After added dedicated widget troubleshooting section and removed confusing help for implementation of custom widget
LGTM

PR #2179, opened 2022-10-05
feat(archives): keep a single paragraph for 7.6 and 7.7

tbouffard:
We had a paragraph for Bonita BPM from 7.3 to 7.5, to group version prior the Bonita rename that occurred in v7.6.
In 7.7, this is still the same product name and archive structure, so 7.6 and 7.7 must be stored in the same paragraph.


benjaminParisel (approved):
:exploding_head:

PR #2184, opened 2022-10-07
ci: add write permissions for the stale workflow

tbouffard:
We are going to set default permissions to read. So use explicit write permissions when needed.

Also bump actions/stale from v4 to v6

covers https://github.com/bonitasoft/bonita-documentation-site/issues/443

PR #2185, opened 2022-10-07
ci: add write permissions for the check PR title workflow

tbouffard:
We are going to set default permissions to read. So use explicit write permissions when needed.

See also #2184
covers https://github.com/bonitasoft/bonita-documentation-site/issues/443


tbouffard:
✔️ logs of the workflow run

```
GITHUB_TOKEN Permissions
  Metadata: read
  PullRequests: write
```

benjaminParisel (approved):

PR #278, opened 2017-05-03
fix(bpm-api) fix typos in context api response payload

cpuy:

PR #2190, opened 2022-10-11
fix(sync): make ldap sync update users with different case (#278)

vhemery:
Consider that two entries wich differ only by the case are meant to be the same entry.

Documentation for fix [RUNTIME-278](https://bonitasoft.atlassian.net/browse/RUNTIME-278)


vhemery:
Yep. Message is still there, with less technical details.
Fine by me. Thx Valentina.

vbotnari12 (approved):
LGTM.

PR #2193, opened 2022-10-12
fix(alias): add alias for migration-tool page in update-tools-changelog

danila-m:
Relates to [RUNTIME-1461](https://bonitasoft.atlassian.net/browse/RUNTIME-1461)

PR #2194, opened 2022-10-12
fix(alias): add alias for migration-tool page in update-tools-changelog

danila-m:
Sister PR to https://github.com/bonitasoft/bonita-doc/pull/2193
Since 7.11 is out of support, we don't propagate upward from it, hence the separate commit.

Relates to [RUNTIME-1461](https://bonitasoft.atlassian.net/browse/RUNTIME-1461)


tbouffard (approved):
✔️ LGTM, same change as in #2193

PR #2195, opened 2022-10-14
ci: enable Vale (syntax checker) linter

jeromecambon:
The Vale linter checks the doc content and detect typos, incorrect syntax, and provide suggestions. See https://vale.sh for details.
Check [the video](https://github.com/bonitasoft/bonita-documentation-site/issues/162#issuecomment-1279060225) included in the issue for a demo!
Covers https://github.com/bonitasoft/bonita-documentation-site/issues/162
See also #2196


.github/workflows/contribution-check.yml
```diff
@@ -17,9 +15,23 @@ jobs:
...
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        diffDoesNotContain: '["https://documentation.bonitasoft.com/", "Bonita BPM"]'
-        extensionsToCheck: '[".adoc"]'
+      - name: Check PR content
+        uses: bonitasoft/actions/packages/pr-diff-checker@1.1.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          diffDoesNotContain: '["https://documentation.bonitasoft.com/", "Bonita BPM"]'
+          extensionsToCheck: '[".adoc"]'
+      - uses: actions/checkout@v2
```
tbouffard, line 24: let's use v3

.github/workflows/contribution-check.yml
```diff
@@ -17,9 +15,23 @@ jobs:
...
+
+      - id: files
```
tbouffard, line 25:
```suggestion
      - id: files
        name: Detect changed files
```

.github/workflows/contribution-check.yml
```diff
@@ -17,9 +15,23 @@ jobs:
...
+        uses: jitterbit/get-changed-files@v1
+        with:
+          format: 'json'
+
```
tbouffard:
```suggestion
```

.github/styles/Vocab/Base/accept.txt
```diff
@@ -0,0 +1,9 @@
+Bonitasoft
+Bonita
+Asciidoc
```
tbouffard: ⚠️ wrong syntax, see https://docs.asciidoctor.org/asciidoc/latest/
```suggestion
AsciiDoc
```

tbouffard (changes requested):

.github/workflows/contribution-check.yml
tbouffard:
```suggestion
      - uses: actions/checkout@v3
```

tbouffard: already applied

tbouffard:
The dictionary is now OK. I don't have time to review the rules today. I wonder if there are any configured rules that could generate false positives.
Let's wait for someone else to review.

jeromecambon:
> The dictionary is now OK. I don't have time to review the rules today. I wonder if there are any configured rules that could generate false positives.
> Let's wait for someone else to review.

The "dictionary" (accept.txt file) will evolve with time, it's just a start. We'll add new keywords when needed.
And yes, we'll have false positive. Typically when the page includes variable names, we can have many. So don't expect the check will be always green...

benjaminParisel (approved):
LGTM :clap:

tbouffard (approved):
> And yes, we'll have false positive. Typically when the page includes variable names, we can have many. So don't expect the check will be always green...

OK, that's clear. False positives are a pain and if we have too many, nobody will look at the results of the checks. Let's start with what is proposed here and see how it runs.

tbouffard:
Put in Draft as we are waiting for feedback from the Product Team.

PR #2189, opened 2022-10-10
fix(2021.1): remove hard coded links to the production documentation site

vbotnari12:
- remove hard coded links to documentation.bonitasoft.com and use xref instead
- adding bcd branch when building the preview to make xref validation pass

PR #2021, opened 2022-03-30
chore(release): archive version 7.6

baptistemesta:

PR #2201, opened 2022-10-21
fix: remove hardcoded links to BCD, Cloud and Bonita docs

tbouffard:
Use xref instead of production urls. This allows xref validation. Update the 'Build PR Preview' workflow as the Bonita 2021.1 version is now needed for xref validation

### Notes

This fix was applied to 2021.1 with #2189 but the merge didn't work well. Some pages moved in 2021.2 but Git lost the links with the original pages (not a real git move, but a delete/add operation).
This PR also includes pending fixes in 2021.1 done in #2200


benjaminParisel:
Tested and it's work well.
Howewer, do you know with some link (cf first screenshot) are not display as classic link (with blue) ?

image


In same page, we got this style for others links

image

benjaminParisel (approved):

tbouffard:
> Tested and it's work well. Howewer, do you know with some link (cf first screenshot) are not display as classic link (with blue) ?

I guess you are talking about this page https://bonitasoft-bonita-doc-build_preview-pr-2201.surge.sh/bonita/2021.2/portal-removal

For the first screenshots, the link are resolved while building the preview as the links refer to page available in the 2021.2 version (the version we are updating).
`<a href="release-notes#_new_bonita_applications" class="xref page">converted into Living Applications</a>`

For the second screenshots, the links are not resolved as they target version 2021.1 and this version is not available in the preview.
`<a href="#2021.1@ROOT:analytics.adoc" class="xref unresolved">Analytics</a>`

The `unresolved` CSS class set a custom color, so I guess this is why the display is not the same
```css
.doc a.unresolved {
  color: var(--link_unresolved-font-color);
}
```

To correctly check the display of links to other vresions/components, we should check the site built by the build-pr-preview workflow. No artifact is available for now to test locally, but this is something we could change.

PR #2206, opened 2022-10-26
fix(SAML SSO): Reformat troubleshooting tip.

ugaston:
Versions: 2021.1, 2021.2, 2022.1, 2022.2

Change the troubleshooting tip format from old to the new one.

### DIFF
--- a/modules/ROOT/pages/single-sign-on-with-saml.adoc
+++ b/modules/ROOT/pages/single-sign-on-with-saml.adoc
@@ -389,7 +389,7 @@ There could be several causes for this, but the most likely are:
 [.troubleshooting-section]
 --
 [.symptom]
-SAML fails after authenticating succesfully in the SSO login page.
+SAML fails after authenticating successfully in the SSO login page.
 
 [.symptom-description]
 The following stacktrace appears in the Bonita server log :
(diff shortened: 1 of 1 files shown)


cvidaillac (changes requested):
It should be nice to have all troubleshooting sections in same new format ;-) !

benjaminParisel:
Nice start! :muscle:

modules/ROOT/pages/single-sign-on-with-saml.adoc
```diff
@@ -362,103 +363,210 @@ org.bonitasoft.level = FINE
...
-*Possible Solutions:*
+[.cause]#Potential cause#
+Some IdPs dissociate the Recipient URL from the Assertion Consumer Service (ACS) URL in the SAML service configuration.
+
+[.solution]#Possible solution#
+The workaround for this issue is to make sure to set the Recipient parameter in the IdP configuration with the same value as the ACS of the service provider (`http[s]://<bundle host>:<port>/bonita/saml`), so that the `destination` attribute is present in the SAML response. This issue hapens when the Recipient URL is left empty resulting to an empty `destination` attribute in the SAML response.
```
abirembaut:
```suggestion
The workaround for this issue is to make sure to set the Recipient parameter in the IdP configuration with the same value as the ACS of the service provider (`http[s]://<bundle host>:<port>/bonita/saml`), so that the `destination` attribute is present in the SAML response. This issue happens when the Recipient URL is left empty resulting to an empty `destination` attribute in the SAML response.
```

abirembaut (approved):
Nice !
just one small (already existing) typo to fix

PR #2215, opened 2022-11-04
ci: make the "Vale action" use the latest release

jeromecambon:
The Vale action was set to 'reviewdog' branch by error. It is not stable and the action failed to run from time to time.

In addition, restore permissions for the 'pr-title' job that was removed by mistake in c2e10ec6.


tbouffard (approved):
LGTM, workflow run pass and uses `errata-ai/vale-action/v2.0.1`
Note that we could have use the v2 tag for auto update

PR #2218, opened 2022-11-07
ci: disable vale-action

tbouffard:
It marks the workflow as failed whereas it is set to provide help to contributors.
All Pull Requests are currently marked in error: disable the action until we find a way to fix the problem/

covers https://github.com/bonitasoft/bonita-documentation-site/issues/464


tbouffard:
ℹ️ received go from @benjaminParisel and @jeromecambon by private message, so merging.

PR #2222, opened 2022-11-10
docs: use the new TS format in bonita-bpm-studio-installation

cvidaillac:
Please find updates on the troubleshooting section for bonita-bpm-studio-installation doc according to the new troubleshooting  format.


rbioteau (approved):
LGTM!

PR #2223, opened 2022-11-10
docs(docx connector): Use new format for troubleshooting section

cvidaillac:
Update done according to the new troubleshooting format proposed by R&D.


cvidaillac:
I did a change on version 2022.2 version instead 2021.1.
Is there a way to move this change in 2021.1 branch or something like that or we need to reject this change and then redo the work based on 2021.1 branch ?

benjaminParisel:
> I did a change on version 2022.2 version instead 2021.1. Is there a way to move this change in 2021.1 branch or something like that or we need to reject this change and then redo the work based on 2021.1 branch ?

Is it always accurate? I see 2021.1.

You need to follow the standart about pull request title (you can check the comments from github-actions-bot for more details). To edit your title, go on the top, and click on `edit button` on the top-right corner

vbotnari12 (approved):
LGTM

PR #2228, opened 2022-11-16
docs(setup tool): remove duplicated descriptions

jeromecambon:
Keep only one header description instead of 3.


tbouffard (approved):
Nice catch!


before | now
----- | -----
image | image

PR #2229, opened 2022-11-18
ci: enable vale-action with neutral state

jeromecambon:
Added the correct parameters so that the action returns a "neutral" state, even in case of errors (or warnings)

Covers [464](https://github.com/bonitasoft/bonita-documentation-site/issues/464)


benjaminParisel (approved):
LGTM

PR #2254, opened 2022-11-29
docs(uid_variables): Call a Rest API on a condition

jeromecambon:
Covers [UID-715](https://bonitasoft.atlassian.net/browse/UID-715)

### DIFF
--- a/modules/pages-and-forms/pages/variables.adoc
+++ b/modules/pages-and-forms/pages/variables.adoc
@@ -22,7 +22,8 @@ There are several types of variables:
 
 String and JSON variables are initialized at page data model instantiation and their values do not change subsequently.
 
-Expression, external API, and URL parameter variables are evaluated every time something happens on the page, such as when receiving data or on a user interaction. The call of an External API variable is triggered every time the URL of the resource it points at changes.
+Expression, external API, and URL parameter variables are evaluated every time something happens on the page, such as when receiving data or on a user interaction.
+The call of an External API variable is triggered at page load, and every time the URL of the resource it points at changes.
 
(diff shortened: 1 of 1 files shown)


pabloalonso (commented):
If we do it in this way, there will be an issue as the REST API permissions will not be set. This may require a second action to update the file page.properties which is not native in UID.

The other way is to do this way:
../API/bpm/userTask/{{callTask}}
This will trigger the call anyway with 'undefined' probably (to double check)

pabloalonso:
NOTE: In addition another use case that we need to cover is how to autorefresh the call with a t={{timestamp}}

jeromecambon:
> If we do it in this way, there will be an issue as the REST API permissions will not be set. This may require a second action to update the file page.properties which is not native in UID.
> 
> The other way is to do this way: ../API/bpm/userTask/{{callTask}} This will trigger the call anyway with 'undefined' probably (to double check)

Good catch!
The latest is ok. An External API variable does not trigger a call in case of undefined variable in the URL.

jeromecambon:
> 

I suggest to handle this in a separate PR. Can you create a Jira for this?

jeromecambon:
> 

I suggest to handle this in a separate PR. Can you create a Jira for this?

jeromecambon:
> NOTE: In addition another use case that we need to cover is how to autorefresh the call with a t={{timestamp}}

I suggest to handle this in a separate PR. Can you create a Jira for this?

pabloalonso:
LGTM

pabloalonso (approved):

PR #2255, opened 2022-11-29
docs: Use environment variable

benjaminParisel:
* Sometimes when we create a custom widget, we didn't see the template in the editor
* Adding a troubleshooting section to explain the cause and the solution


vbotnari12 (approved):
LGTM

PR #2268, opened 2022-12-05
docs(Rest API refresh): Call a Rest API regularly (auto refresh)

jeromecambon:
Covers [UID-717](https://bonitasoft.atlassian.net/browse/UID-717)


benjaminParisel:
Why not on the 2021.1 branches the latest supported version?

jeromecambon:
> Why not on the 2021.1 branches the latest supported version?

To avoid merge, because of 2021.2 re-structure. Since it is a minor update and in few months 2021.2 will not be supported, this is not an issue IMHO.

benjaminParisel:
Miss click sorry for the close x) 

I suggest you to target this PR on 2021.1 to be in adequation with the common rules ...

PR #2278, opened 2022-12-22
docs: Add troubleshooting section on messaging

cvidaillac:
Based on customer case we've got (00027068 - https://bonitasoft.lightning.force.com/lightning/r/Case/5002p00002pUaATAA0/view), we are adding a new troubleshooting section for all customers that may face the same issue.


vbotnari12 (approved):
ok

PR #2279, opened 2022-12-23
ci: Add images compression workflow (test)

jeromecambon:
Testing image compression action


tbouffard:
What is the relation with #2280?
#2279 is targeting the right branch

PR #2280, opened 2022-12-23
ci: Add new compress-images action (+ first launch)

jeromecambon:
The new compress-images workflow is launched on each image added or changed.
An image is modified (compressed) to initiate the first launch.
I have checked that after this first launch, all these images will not be compressed again.

Covers https://github.com/bonitasoft/bonita-documentation-site/issues/434


.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Compress Images
+        uses: calibreapp/image-actions@main
+        with:
+          # The `GITHUB_TOKEN` is automatically generated by GitHub and scoped only to the repository that is currently running the action. By default, the action can’t update Pull Requests initiated from forked repositories.
+          # See https://docs.github.com/en/actions/reference/authentication-in-a-workflow and https://help.github.com/en/articles/virtual-environments-for-github-actions#token-permissions
```
tbouffard: Please remove this comment. It is useless (I know it comes from the  `calibreapp/image-actions` README)

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
...
+      - '**.jpeg'
+      - '**.png'
+      - '**.webp'
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  build:
```
tbouffard: Please use convenient job name. We don't build anything here.

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
...
+    # Only run on Pull Requests within the same repository, and not from forks.
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
tbouffard: I know this is what is documented in the `calibreapp/image-actions` README, but this is not the best way to detect fork. The `github.event.pull_request.head.repo.fork` property is storing the information.
For instance, see usage in 
- https://github.com/bonitasoft/actions/blob/v2.1.0/packages/pr-title-conventional-commits/action.yml#L52-L57
- https://github.com/afc163/surge-preview/commit/97be9f4b918a0c329d0f6a5d2890626a884fe830#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3d

In addition, the comment must explain why it is not run when the PR is created from a fork. The if condition will be explicit enough, no need to paraphrase the code with a "what" comment.

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
+name: Compress Images
+on:
+  pull_request:
+    # Run Image Actions when JPG, JPEG, PNG or WebP files are added or changed.
+    # See https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths for reference.
```
tbouffard: Useless comment, please remove

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
...
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  build:
+    # Only run on Pull Requests within the same repository, and not from forks.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    name: calibreapp/image-actions
+    runs-on: ubuntu-latest
```
tbouffard: Please use ubuntu-22.04 runner as we do in other workflows. We don't want to depend on GitHub defaults that can change at any time and break think silently.

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
```
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Compress Images
+        uses: calibreapp/image-actions@main
```
tbouffard: ⚠️ please use a specific version. This action seems not providing tag which is bad for many reasons (traceability, security, ....) that have been massively documented by GitHub. Please vote for https://github.com/calibreapp/image-actions/issues/128
In the meantime, we should use a specific commit.

tbouffard (changes requested):
❌ For all the reasons explained below, I strongly veto the merging of this Pull Request.
Please talk with the Team prior opening such Pull Requests: we are loosing time in review and discussions. The topic is not mature enough to be ready for implementation.
We never discuss about introducing such workflow in https://github.com/bonitasoft/bonita-documentation-site/issues/434, we were proposing to add warnings about image size and they were prerequisites on the technical sides to be considered.

The oldest supported version is 2021.1, so the PR should be applied to this branch first, no matter we have merge issue in the 2021.2 branch because of wrongly moved images.

The workflow has to be reworked because it doesn't follow our standards.

I guess @benjaminParisel and @jeromecambon had conversations about the "shared workflow topics", but in its current form, this PR introduces potential configuration duplication. It is against the work we are performing with https://github.com/bonitasoft/bonita-documentation-site/issues/422 and https://github.com/bonitasoft/bonita-documentation-site/issues/472
We already introduce the "vale" action in bonita-doc only. There must be a plan to apply such changes in all repositories, and I don't see any issue to track that.

I am not sure we want to automate such images transformations, this should be discuss and we should list pros/cons.

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,31 @@
+name: Compress Images
+on:
+  pull_request:
+    # Run Image Actions when JPG, JPEG, PNG or WebP files are added or changed.
+    # See https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths for reference.
+    paths:
+      - '**.jpg'
```
tbouffard, line 6: also add the workflow file to ensure it is run when it changes. We always do this with workflows

.github/workflows/compress-images.yml
jeromecambon: done

.github/workflows/compress-images.yml
jeromecambon: done

.github/workflows/compress-images.yml
jeromecambon: Removed 2nd comment. First one is useful IMO.

.github/workflows/compress-images.yml
jeromecambon: done

.github/workflows/compress-images.yml
jeromecambon: done

.github/workflows/compress-images.yml
jeromecambon: done

.github/workflows/compress-images.yml
jeromecambon: done

.github/workflows/compress-images.yml
jeromecambon: Done

jeromecambon:
After discussion, we decided to:
- Fix the PR comments
- Do a PR on 2021.1 and remove this one
- Apply a "manual" compression of images on older versions

PR #2289, opened 2023-01-09
docs: update broken link for messaging

benjaminParisel:


benjaminParisel:
The error on the build comes from the 2022.2 branch. This PR will be fixed it after running the `moulinette` action

PR #2291, opened 2023-01-09
refactor(2021.2): compress images

tbouffard:
The images were initially compressed in the `2021.1` branch. Some changes haven't been applied to the `2021.2` branch because of merge conflicts (some images have moved in the `2021.2` and git isn't able to detect all moves). So, run the compression tool again in `2021.2`.


.github/workflows/compress-images.yml
```diff
@@ -1,3 +1,4 @@
+# TODO remove this comment - already created to trigger a workflow run
```
```suggestion
```

PR #2297, opened 2023-01-17
fix(Manual operations): simplify documentation

JeremJR:
simplify documentation as for these 3 files the only differences would be in comments


vbotnari12 (approved):
ok :)

PR #2316, opened 2023-02-09
docs(info): Update rest-api-authorization typo err

cvidaillac:
Typo error in the Compound permissions mapping section. 
By default, the file, the name of the file is "compound-permissions-mapping.properties" not "resources-permissions-mapping.properties". 
Ideally, in this section, we could update it with use cases section to demonstrate some permissions capabilities.


ugaston (approved):
Good catch!
Looks good to me.

PR #2321, opened 2023-02-13
fix prometheus metrics names

laurentleseigneur:
FIX[prometheus]: documented metrics name was falsy

* add support for ...
* fix ... behaviour
* refactor ....


passga (changes requested):
It's only the prometheus endpoint which replace all `.` all metrics by the underscore `_` for the metrics name.  If you choose to display a metrics in logs the metrics will be logged like this 

`bonita_1    | {"instant":{.... logger","level":"INFO","loggerName":"com.bonitasoft.engine.monitoring.LoggingMeterRegistry","message":"bonita.bpmengine.connector.pending {tenant=1} ....}`

And i think it will also the case in **jmx** and with the **Spring Boot** endpoint  actuator ( comming soon ). 

So  just mentione in **Prometheus** section that the metrics will be displayed with the specific **Prometheus** format.

laurentleseigneur:
@passga since it's official documentation about how to use prom metrics, when I want to enable prometheus metrics and use them in a Grafana dashboard, it's the `/metric` endpoint and metrics exposed names that should be used in documentation
IMO doc should explain this, since i needed to update metric names when I pushed them to prom/grafana stack

laurentleseigneur:
up on this PR

benjaminParisel:
@laurentleseigneur Are you sure to merge this PR on 7.11 branch? If yes, be careful, this changes will not be propagate to others version automatically

laurentleseigneur:
it's supposed to be the first version that introduce this page
@benjaminParisel ,  should I use first supported version instead?

benjaminParisel:
> it's supposed to be the first version that introduce this page @benjaminParisel , should I use first supported version instead?

Yes, use 2021.2 :)

passga:
> up on this PR

As i said , i'm  not ok to merge this PR with the actual content. Only the part [Prometheus](https://documentation.bonitasoft.com/bonita/2022.2/runtime/runtime-monitoring#_prometheus_publisher.) should be changed to give more details i it needed. So @laurentleseigneur  if you have time, you could update your pr to add the new information's if you want.

laurentleseigneur:
replaced by #2358

PR #2324, opened 2023-02-16
fix(asciidoc): workaround wrong Maven pom xml display

educhastenier:
Works around issue https://github.com/asciidoctor/asciidoctor/issues/3128

closes [RUNTIME-969](https://bonitasoft.atlassian.net/browse/RUNTIME-969)
closes [RUNTIME-1652](https://bonitasoft.atlassian.net/browse/RUNTIME-1652)

[RUNTIME-969]: https://bonitasoft.atlassian.net/browse/RUNTIME-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUNTIME-1652]: https://bonitasoft.atlassian.net/browse/RUNTIME-1652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/ROOT/pages/custom-authorization-rule-mapping.adoc
+++ b/modules/ROOT/pages/custom-authorization-rule-mapping.adoc
@@ -37,7 +37,7 @@ In this example, Custom authorization rule is a maven-based java project that ne
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="\http://maven.apache.org/POM/4.0.0 \http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>org.bonitasoft.example</groupId>
--- a/modules/ROOT/pages/event-handlers.adoc
+++ b/modules/ROOT/pages/event-handlers.adoc
@@ -64,7 +64,7 @@ pom.xml :
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+        xsi:schemaLocation="\http://maven.apache.org/POM/4.0.0 \http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>org.bonitasoft.example</groupId>


modules/ROOT/pages/customize-living-application-theme.adoc
```diff
@@ -134,7 +134,7 @@ The maven descriptor is responsible for running the npm build and package the re
 ----
 <project xmlns="http://maven.apache.org/POM/4.0.0"
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	xsi:schemaLocation="\http://maven.apache.org/POM/4.0.0 \http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
tbouffard: **issue (blocking):** the page displays extra `\`

image

modules/ROOT/pages/event-handlers.adoc
```diff
@@ -61,10 +61,9 @@ pom.xml :
 // for the 'subs' parameter, see https://docs.asciidoctor.org/asciidoc/latest/subs/apply-subs-to-blocks/
 [source,xml,subs="+macros"]
 ----
-<?xml version="1.0" encoding="UTF-8"?>
```
tbouffard, line 64: **question:** why removing the xml directive?

educhastenier: it was a test that I needed to revert and I forgot.
Will revert it.

modules/ROOT/pages/customize-living-application-theme.adoc
educhastenier: reverted for this file, but I do NOT understand why we don't have the same problem here 😞 .

modules/ROOT/pages/event-handlers.adoc
educhastenier: done

modules/ROOT/pages/custom-authorization-rule-mapping.adoc
```diff
@@ -37,7 +37,7 @@ In this example, Custom authorization rule is a maven-based java project that ne
 <?xml version="1.0" encoding="UTF-8"?>
```
tbouffard, line 37: **thought:** I have just discovered that we have the issue only when `<?xml version="1.0" encoding="UTF-8"?>` is present before the line where we see the issue.
I have detected this after you revert a change including the workaround that is not working: https://github.com/bonitasoft/bonita-doc/pull/2324#pullrequestreview-1301658283

Did you try to remove the XML directive to see if this doesn't fix the problem without the need of the workaround?

educhastenier: Yes, that was my motivation, but it did not seem to be enough to get rid of the problem.

tbouffard: OK, fine. 😸

tbouffard (approved):
LGTM, good job for a painful issue 😈

PR #2389, opened 2023-05-12
chore(vale): Cleanup Vale syntax checker (not used)

jeromecambon:
Vale syntax checker issue is not fixed for a while, so we need to cleanup Vale-related info.
See https://github.com/bonitasoft/bonita-documentation-site/issues/464


tbouffard (approved):
The .github/styles/Vocab/Base should be removed as well as it contains the Vale configuration.
Otherwise, LGTM

PR #2391, opened 2023-05-15
fix(docker): fix java_opts to reflect what is in the code

4rthurLegrand:
* fix JAVA_OPTS parameter to reflect what is really in the code


passga (approved):
LGTM!!

PR #2407, opened 2023-05-24
docs(RN): warning regarding experimental packaging tool

vbotnari12:
since the update process will not be covered by Bonitasoft, a warning to deploy apps only in non-production env has to be added.


rbioteau (approved):

akantcheff (approved):
LGTM

PR #2414, opened 2023-06-16
docs(README): fix typo in AsciiDoc

tbouffard:
AsciiDoc is a trademark and case matters.


tbouffard:
The "Contribution checks" workflow reports an error but it should be ignored. See https://github.com/bonitasoft/actions/issues/112

PR #2417, opened 2023-06-23
ci: use ubuntu 22.04 in all workflows

tbouffard:
Some workflows were still using ubuntu 20.04


tbouffard:
Note that an extra comment listing the undefined/undefined url is created. This is a bug which is tracked in https://github.com/bonitasoft/bonita-documentation-site/issues/534

benjaminParisel (approved):

PR #2425, opened 2023-07-07
fix(asciidoctor): fix warning logged at build time

tbouffard:
The warnings were due to issues with xref. Also fix info logs due to wrong xref.
This means that there is no more `INFO (asciidoctor)` or `WARN (asciidoctor)` in the build logs.

The workflow that checks changes now prohibits certain erroneous forms of xref and link, to prevent new errors from being added.

### Tasks

- [x] fix the warnings
- [x] when https://github.com/bonitasoft/bonita-documentation-site/pull/596 is merged, update the workflow to use the master branch of the documentation-site repository


benjaminParisel (approved):
nice job ! :clap:

PR #2430, opened 2023-07-21
docs(RN): add warning about uploaded files and database storage

abirembaut:


modules/ROOT/pages/release-notes.adoc
```diff
@@ -25,7 +25,8 @@ When configuring a xref:runtime:overview-of-bonita-bpm-in-a-cluster.adoc[Bonita
 
 === REST API and file upload
 
-When using the end point portal/fileUpload to upload a file before using it in Bonita REST API or a REST API extension, the size of the file that can be uploaded is now limited to 25MB. This can be changed in the configuration file `console-config.properties` using the setup tool, but be aware that this will affect upload limit for all users.  
+When using the end point `portal/fileUpload` to upload a file (in order to then use it in Bonita REST API or a REST API extension), the size of the file that can be uploaded is now limited to 25MB. This can be changed in the configuration file `console-config.properties` using the setup tool, but be aware that this will affect upload limit for all users (upload end points used for administration like process upload etc... are not affected by this limit). +
+Using Bonita Runtime upload end points for large files (hundreds of MB) is not recommended as the database is used to store temporarily those files (and there can be a 1 or 2GB per file limit depending on the database vendor). Prefer using an external system if you need to manage very large files in your applications.
```
educhastenier, line 29: > as the database is used to store temporarily those files

Isn't it too much technical information that an attacker could try to exploit? I mean, do we **need** to precise that the content is going to end up into the database?

abirembaut: maybe... @pabloalonso @tha-bt what is your opinion ?

abirembaut: otherwise we could say, with or without the "(depending on the database vendor)" :
```suggestion
Using Bonita Runtime upload end points for large files (hundreds of MB) is not recommended as the temporary storage for those files can have a 1 or 2GB per file limit (depending on the database vendor). Prefer using an external system if you need to manage very large files in your application.
```

PR #2450, opened 2023-09-26
docs: correct spelling 

triguimed:
an icon not a icon

PR #2452, opened 2023-09-26
docs spell error ( an icon not a icon) 

triguimed:
docs an icon not a icon


tbouffard (changes requested):
Duplicate of #2450, must be done on the 2021.2 branch

PR #2464, opened 2023-10-16
docs(relaseNote): update release notes and add glossary page

rbioteau:


modules/ROOT/pages/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+Organization::
+    The definition of the user base that will use the processes (includes users, roles and groups).
+
+== P
+
+Pages::
+    Web pages used within your applications.
+
+Project::
```
jeromecambon: It is surprising to have so much details on this. IMO we should have a simple description here, with these deatils moved in the `workspaces-and-repositories` page.

akantcheff (approved):

PR #2455, opened 2023-10-05
docs(sca): documentation related to the new SCA mode

rbioteau:
Technical documentation on how to use the maven build.

PR #2485, opened 2023-10-24
docs(SCA): add note on authentication modes not SCA compliant

abirembaut:
The following authentication mode are not fully SCA compliant :
- Kerberos SSO
- CAS SSO
- LDAP auth

* Add note at the top of the page

### DIFF
--- a/modules/identity/pages/active-directory-or-ldap-authentication.adoc
+++ b/modules/identity/pages/active-directory-or-ldap-authentication.adoc
@@ -9,6 +9,11 @@
 For Subscription editions only.
 ====
 
+[NOTE]
+====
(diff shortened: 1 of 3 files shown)


modules/identity/pages/active-directory-or-ldap-authentication.adoc
```diff
@@ -9,6 +9,11 @@
 For Subscription editions only.
 ====
 
+[NOTE]
+====
+The configuration steps detailed in this page include some modifications to the Tomcat configuration. This is not compliant with the Self Contained Application deployment mode introduced with Bonita 2023.2. If you want your Self Contained Application to work with an Active Directory/LDAP authentication, you will need to build the docker image of the SCA first and then configure the container (using a volume for example).
```
akantcheff: Self Contained Application or SCA is not the commercial name. It is "Application" and "Application Mode".

Can you change the target branch to #2455 so we can regroup all modifications related to SCA topics? So the notes can link to the new introduced mode.

abirembaut: @akantcheff done

PR #2487, opened 2023-10-25
feat(configuration): property source order

rbioteau:


modules/ROOT/pages/release-notes.adoc
```diff
@@ -304,6 +304,10 @@ With the new Self-Contained Application (abbr. SCA) approach, creating or updati
 A new `bonita.runtime.startup.update-only` configuration boolean property has been added to perform an update only startup.
 When enabled, the application will start, go through its initialization and update phase and exit afterward. Exit status can be successful (`0`) or in error (`1`).
 
+=== Properties source priority order
+
+A new `bonita.runtime.properties.order.legacy-mode` configuration boolean property has been added. When enabled, properties files from database have a higher priority order over environment and system variables. By default, this property is disabled and properties from the database can be overridden with environment or system variables.
```
akantcheff: suggestion:
I think we should put in front that the main changement here is the priority order rather than this new property (that is a result of keeping a legacy mode that we will remove afterward).

```suggestion
Prior Bonita 2023.2, properties files from database had priority over environment or system variables. We decide to reverse this order so that you can easily override database properties with environment or system variables, and to be in line with the https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config[Spring / Spring Boot philosophy].

To be backward compatible, we introduce the boolean property `bonita.runtime.properties.order.legacy-mode` (default: `false`) to enable the previous priority order of properties source.
```

source: https://github.com/bonitasoft/bonita-engine-sp/pull/2717/files#diff-47a8d24df76414b9397c7d361743c06b4b8c0558a4ae76ff815f09b9a164f061

akantcheff (approved):

PR #2488, opened 2023-10-26
fix(releaseNote): property update

rbioteau:


educhastenier (approved):
Looks good to me :+1:

PR #2518, opened 2023-11-22
docs: Update format error

benjaminParisel:
We don't want to display this example as an adoc array.


benjaminParisel:
Before:
image

After: 
image

rbioteau (approved):

PR #2556, opened 2023-12-13
feat(properties): document how to configure Bonita with properties

educhastenier:
* with Setup Tool
* via Java Sysprops
* via Env vars

### DIFF
--- a/modules/runtime/pages/configuring-bonita-with-properties.adoc
+++ b/modules/runtime/pages/configuring-bonita-with-properties.adoc
@@ -1,5 +1,6 @@
 = Bonita Runtime configuration
 :description: This page explains how Bonita runtime can be configured through properties, via database, Java System properties, or environment variables
+:tabs-sync-option:
 
 {description}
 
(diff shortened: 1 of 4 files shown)


modules/runtime/pages/configuring-bonita-with-properties.adoc
```diff
@@ -0,0 +1,179 @@
...
+      - BONITA_RUNTIME_PROPERTIES_ORDER_LEGACY_MODE=true
+# ...
+----
+
+
+== limitations
+
+Some properties cannot be changed via System Properties or Environment Variables:
+
+* TODO
```
vhemery: No TODO in the final doc site... 😉

vhemery (changes requested):
I can't find the configuring-bonita-with-properties.adoc from the navigation menu.
It should be added in runtime-configuration-index and in runtime-nav.adoc

Also see suggestions inline.

vhemery:
> I can't find the configuring-bonita-with-properties.adoc from the navigation menu. It should be added in runtime-configuration-index and in runtime-nav.adoc
> 
> Also see suggestions inline.

Argh.. my most meaningfull comment was lost...
I was suggesting to use alternative presentation for:

- In a Linux Shell:
- When starting a Bonita Docker container:
- In a docker-compose.yml file:
See https://github.com/bonitasoft/bonita-doc/edit/feat/properties/document_bonita_properties/modules/runtime/pages/bonita-docker-installation.adoc and search for `[tabs]` for examples.

modules/runtime/pages/configuring-bonita-with-properties.adoc
educhastenier: Fixed in next commit

educhastenier:
> > I can't find the configuring-bonita-with-properties.adoc from the navigation menu. It should be added in runtime-configuration-index and in runtime-nav.adoc
> > Also see suggestions inline.
> 
> Argh.. my most meaningfull comment was lost... I was suggesting to use alternative presentation for:
> 
> * In a Linux Shell:
> * When starting a Bonita Docker container:
> * In a docker-compose.yml file:
>   See https://github.com/bonitasoft/bonita-doc/edit/feat/properties/document_bonita_properties/modules/runtime/pages/bonita-docker-installation.adoc and search for `[tabs]` for examples.

Feedbacks integrated.
Thx for the suggestions.

vhemery (approved):
Looks perfect to me

PR #2567, opened 2024-01-25
feat(java17): update with Java 17

vhemery:
Information about package opening when updating to Java 17


vhemery:
Not sure why the title level is different after Update case overview pages before updating to 7.8.0...
Maybe it's worth doing a separate to level section for Java updates...

educhastenier (approved):

vhemery:
Modifications are already available in the Update Tool. No need to document in the old Migration Tool.

PR #2600, opened 2024-02-12
fix(duplicate): Update update-with-migration-tool.adoc

techosbonitasoft:
remove JRE requirements duplicate section


vhemery (changes requested):
The change is correct, but the base branch is not the correct one and techosbonitasoft account should not be used for lack of tracability.

I'll create an alternative PR with this fix.

vhemery:
Fix contributed by https://github.com/bonitasoft/bonita-doc/pull/2603 instead

PR #2615, opened 2024-02-19
fix(build): merge-configuration clarification

vhemery:


modules/build-run/pages/build-application.adoc
```diff
@@ -172,7 +172,9 @@ It is possible to override an existing file using the `-Dparameters.overwrite` s
 
 === Merge parameters configuration
 
-During the build of the application the `merge-configuration` goal is executed and the content of the `parameters-${bonita.environment}.yml` file is merged into the final Bonita configuration file.
+During the build of the application the `merge-configuration` goal (https://bonitasoft.github.io/bonita-project-maven-plugin/merge-configuration-mojo.html[bonita-project:merge-configuration]) is executed automatically.
+
+The content of the `parameters-${bonita.environment}.yml` file located in `.bcd_configurations` folder of your project is merged into the final Bonita configuration file.
```
educhastenier, line 177: This is weird to talk about BCD here whereas we did not talk about it before.
At least it should be introduced.

educhastenier: Moreover, it should be stated that this part is for Subscription only

educhastenier (changes requested):
2 remarks

modules/build-run/pages/build-application.adoc
vhemery: > Moreover, it should be stated that this part is for Subscription only

It's not... bonita-project-maven-plugin is public.
https://github.com/bonitasoft/bonita-project-maven-plugin/blob/c05767b60e00c0768f435915ec300a942491230c/plugin/src/main/java/org/bonitasoft/plugin/build/AbstractConfigurationArchiveMojo.java#L45
`.bcd_configurations` is just a folder name, which was chosen for compatibility to ease the migration to users of the old BCD product. But BCD is no longer a product, only an approach.

laurentleseigneur: this also impies we can't anymore use the extract conf to a display content of a .bconf file in a readable yaml file

vhemery: @rbioteau Insights about all this ?

rbioteau: > this also impies we can't anymore use the extract conf to a display content of a .bconf file in a readable yaml file

It is not the purpose of the `bconf` file. You may still extract them to retrieve its content as they are just zip file but this is internal implementation and subject to changes without notice. 

@laurentleseigneur If you have feedback to share on how you use or want to use the bconf, please contact the product team.

laurentleseigneur:
any update before this PR will die ?

PR #2641, opened 2024-03-22
docs: add information about deploying bonita on k8s

marcobonita:
New page about deploying Bonita on K8S


educhastenier (commented):
not sure 2024.1 is the right target branch.
I would have put it in 2022.1 instead.

PR #2645, opened 2024-03-27
fix(troubleshoot): add section on how to handle transaction isolation

danila-m:
Relates to [RUNTIME-1830](https://bonitasoft.atlassian.net/browse/RUNTIME-1830)
Engine PR -> https://github.com/bonitasoft/bonita-engine-sp/pull/2911

[RUNTIME-1830]: https://bonitasoft.atlassian.net/browse/RUNTIME-1830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tbouffard:
⚠️ The default branch is now `2022.2`, this branch must target 2022.2

modules/runtime/pages/performance-troubleshooting.adoc
```diff
@@ -102,6 +102,28 @@ The execution threshold value can be configured in `bonita-platform-community.pr
...
+----
+"Within the same transaction, the Search count & page results are not consistent."
+----
+For your information, when Bonita performs a search to build a list of elements (e.g. the Bonita applications listing the process definitions, process instances, human tasks, ...) it always performs two queries on the DB:
+
+ - a first query to count the TOTAL amount of elements;
+ - a second query to actually retrieve the first PAGE_SIZE elements.
+
+If the TOTAL number of elements is smaller than the PAGE_SIZE, Bonita then compares the TOTAL count against the number of elements retrieved by the second query.
+Depending on the number of elements in the DB, those queries may take a long time to execute, thus some update may take place while performing the first query.
```
educhastenier, line 118: This is not true. In this precise case, the problem is that the count query and the search query do no match exactly, so this is a dev problem. The query has to be fixed.

educhastenier (changes requested):
There is a confusion between 2 different detection on the search mechanism.
Let's talk about it.

danila-m:
Closing this PR
As discussed it with the relevant parties, some of the info here is incorrect, and the ticket has to be treated in another way.

PR #2651, opened 2024-04-02
feat: add Bonita 7.10 to archives

educhastenier:


tbouffard (approved):
✔️ download link is OK

PR #2683, opened 2024-05-06
docs(repporting app): fix subscription editions

abirembaut:


educhastenier (approved):
Looks good to me :+1:

PR #2702, opened 2024-05-21
feat(UID2024): Initialize the documentation structure

julienmege:
Relates to https://bonitasoft.atlassian.net/browse/PROD-265


marcobonita:
Few comments:

- Can we rename xref:how-tos-builder.adoc[How-tos] as "Guides" instead? We already have a "how-tos" page in the old UID doc and I wouldn't like it to get confusing 
- Can we move the whole "Forms and pages" section (https://documentation.bonitasoft.com/bonita/latest/pages-and-forms/pages-and-forms) under ref:custom-applications-index.adoc[Living applications with Bonita UID (Legacy)] ?
-  Can remove the "(legacy)" in ref:custom-applications-index.adoc[Living applications with Bonita UID (Legacy)] ? I'm not sure we want to market "legacy" for now

PR #2716, opened 2024-06-06
docs(purge-tool): update release notes

educhastenier:
Also, add the RN for the version 1.0.7 that was missing

### DIFF
--- a/modules/runtime/pages/purge-tool-changelog.adoc
+++ b/modules/runtime/pages/purge-tool-changelog.adoc
@@ -6,6 +6,32 @@ This is the changelog of the purge tool.
 
 The purge tool is used to remove data from bonita archive tables. It is useful for big production environments.
 
+
+== 1.0.8 - June 5, 2024
+
+Now the tool **MUST** be run from the **root folder** by calling:
+
+```shell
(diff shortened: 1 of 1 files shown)


educhastenier:
@cvidaillac can you review this PR please?

cvidaillac:
@educhastenier 2 comments from Support team:
1) We would like to keep historical changes done on the tool by adding fixes done on 1.0.4, 1.0.5, 1.0.6 and 1.0.7 versions.
I can help if you can tell me how I can see the fix details somewhere.
2) Regarding what it has been written for 1.0.8, it would be better to keep same format than other change logs done with a Fixes section and Other section.

educhastenier:
> 1. We would like to keep historical changes done on the tool by adding fixes done on 1.0.4, 1.0.5, 1.0.6 and 1.0.7 versions.
>    I can help if you can tell me how I can see the fix details somewhere.

Do you mean you prefer to keep all versions in the RN, even if the binaries of old versions are removed from download?

cvidaillac (approved):
I'm fine with this version.
Keeping a list of features or bugs through different versions we delieverd is interesting.
Note that some customer may use old version and getting the all changes log list for every versions could make sense. Thanks

PR #2739, opened 2024-06-17
docs: Move in production with bonita-ui-builder

benjaminParisel:
Explain how to go in production and the user flow with bonita-ui-builder

PR #2760, opened 2024-07-04
feat: update UI builder version

vhemery:


antora.yml
```diff
@@ -16,6 +16,7 @@ asciidoc:
     bcdDocVersion: '4.0'
     testToolkitVersion: '3.0'
     openApiLatestVersion: '0.0.18' # latest compatible version of bonita openApi with this Bonita version
+    bonitaUIBuilderVersion: '1.0.0-beta.1'
```
benjaminParisel, line 19: You can remove this attribute, we already add it in another PR #2739

vhemery:
Used latest instead

PR #2785, opened 2024-07-18
docs(applications): split pages into separate folders

akantcheff:


modules/applications/examples/ui-designer/localization.json
akantcheff, line 1: I will create another PR to format this file because if doing so the file is not marked as moved.

marcobonita (approved):
LGTM

benjaminParisel (approved):

PR #2798, opened 2024-07-25
docs:  Adding uib section on guest mode page

benjaminParisel:
* delete the next-steps page in same time (no need to put alias for this page

### DIFF
--- a/modules/identity/pages/guest-user.adoc
+++ b/modules/identity/pages/guest-user.adoc
@@ -59,7 +59,7 @@ Restart Bonita server.
 . If your configuration is correct you should be able to access your public application directly with the URL `http://<host>:<port>/bonita/apps/<app token>` without being logged in first. +
 You are done.
 
-[NOTE.uib,caption=UI Builder]
+[NOTE,caption=UI Builder]
 ====
 If you are using the UI Builder, you need to recreate the UI Builder production container to be able to use the guest user access..
 ====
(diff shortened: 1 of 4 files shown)


benjaminParisel:
> # Contribution Guidelines checks
> The content of the files modified by this Pull Request doesn't match the Contribution Guidelines.
> 
> Please update the following files.
> 
> ## ⛔ Forbidden pattern validation
> Some patterns are forbidden in the following files:
> 
> * [ ]  Update **Bonita BPM** syntax from **modules/identity/pages/guest-user.adoc**

We want to keep this Bonita BPM syntax here to refer the oldest version of Bonita

modules/identity/pages/guest-user.adoc
```diff
@@ -59,6 +59,11 @@ Restart Bonita server.
 . If your configuration is correct you should be able to access your public application directly with the URL `http://<host>:<port>/bonita/apps/<app token>` without being logged in first. +
 You are done.
 
+[NOTE,caption=UI Builder]
+====
+If you are using the UI Builder, you need to recreate the UI Builder production container to be able to use the guest user access..
```
Shmayro: **suggestion (if-minor):** I think it can be a good idea to add the link to the UIB production packaging here

Shmayro: there is double points at the end 👀

Shmayro: A little suggestion.
add "Make sure that your public apps are among UIB packaged apps". no need it's seems already clear. But I see that having something like that could be helpful to the users.

Shmayro (changes requested):
some suggestions

Shmayro (approved):
lgtm

PR #2800, opened 2024-07-26
docs: Document how to avoid dataloss with UIB

vhemery:


modules/applications/pages/ui-builder/download-and-launch.adoc
```diff
@@ -70,6 +70,21 @@ Then, you can access Bonita UI Builder at `http://localhost`.
...
+====
+By default, the database volumes are not persisted. Make sure you export your applications as .json files before running the command.
+[source, shell, subs="+attributes"]
+----
+docker compose down
+----
+If you do not want to loose your data, you can use instead these commands to stop and restart Bonita UI Builder:
+[source, shell, subs="+attributes"]
+----
+docker compose stop
```
laurentleseigneur, line 83: in a beta process, this can be an issue since we want to update from beta1 to beta2 but keep the volume that includes mongo data, I suggest to document which persistent volume is relevant to use  UIb without loose since GIT feature is not yet available

vhemery: That's somthing we will rework, with:

- more modular docker images (which will change the said container)
- the integration with Studio (with more options to save/erase the container)

But for the moment, with the GA, the proposed solution is only to export and reimport all your jsons.
The integration with Studio and this problematic is not part of the GA functional scope.

marcobonita:
Comment to keep it open

marcobonita:
Let's commit the 2 changes I suggested then GO for merge 👍

PR #2853, opened 2024-09-13
docs: Update URL in connector-archetype-tutorial.adoc

nchabanoles:
Use the HTTPS protocol to reach the external service used as example


rbioteau (commented):
Should be done on oldest supported branch

PR #2854, opened 2024-09-13
docs: Update connector-archetype-tutorial.adoc

nchabanoles:
Updated all the URLs toward the external service to use HTTPS


rbioteau (commented):
Should be done on oldest supported branch

PR #2859, opened 2024-09-23
docs: improvements to UIB communication

marcobonita:


modules/applications/pages/ui-builder/faq.adoc
```diff
@@ -141,6 +141,10 @@ Yes, they are compatible with both deployment modes.
 Yes, it is. Reach out to your customer service representative to explore how you can migrate to Bonita Cloud.
 
 
+=== Can Bonita UI Builder be used offline?
```
jeromecambon, line 144: Did you check this? UIB is downloading resources (images, fonts, ...) from remote servers and therefore cannot be used offline. 
Perhaps you launched UIB once online, then restarted offline ? In this case it may have taken the resources from the cache.

modules/applications/examples/ui-builder/development/docker-compose.yml
```diff
@@ -1,10 +1,10 @@
 # Use for running a bonita-ui-builder during development.
 services:
   bonita-ui-builder:
-    image: ${IMAGE_NAME:-pass:a[{bonita-ui-builder-image}]:latest}
+    image: pass:a[{bonita-ui-builder-image}]:latest
```
jeromecambon, line 4: seems weird, from where the image value is taken then ?  did you check this ?

modules/applications/examples/ui-builder/development/docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
 
   # A reverse proxy used to communicate between Bonita UI Builder and a Bonita Runtime
   bonita-ui-proxy:
-    image: ${IMAGE_NAME:-pass:a[{bonita-ui-proxy-image}]:latest}
+    image: pass:a[{bonita-ui-proxy-image}]:latest
```
jeromecambon, line 18: Same comment as above

jeromecambon (changes requested):

modules/applications/examples/ui-builder/development/docker-compose.yml
marcobonita: The ${IMAGE_NAME} was actually hardcoded.
About the 'pass:a' and the image value, I actually don't know where it's coming from. Maybe from Antora variables

When you look at the preview, it gives the expected result:

- before: image: ${IMAGE_NAME:-bonitasoft.jfrog.io/docker/bonita-ui-builder:latest}

- after (see preview): image: bonitasoft.jfrog.io/docker/bonita-ui-builder:latest

modules/applications/pages/ui-builder/faq.adoc
marcobonita: Actually, during my tests, I was able to call different API resources and I even was able to instantiate processes. I couldn't see any big difference with the online mode.

Internet is still mandatory though for the installation (and therefore the first launch) so it's kinda hard to tell whether it's because of cache or not.
I also tried in incognito mode and didn't come across any issues either

modules/applications/examples/ui-builder/development/docker-compose.yml
jeromecambon: Ha ok !
 I just looked at the source file, and yes this is a antora variable. That's fine then.

modules/applications/pages/ui-builder/faq.adoc
jeromecambon: I did some checking: if you empty your cache (Clear site data From dev tools > Storage), and reload your page, you don't get anything. Even without emptying the cache, I noticed side effect (loading delays, ...).

jeromecambon: I did some investigation and discovered the static files (images, fonts, css, ...) are actually stored at build time in the Appsmith server ! So forget my previous comment then.
But we should at least say that some features such as git cannot be used offline.

marcobonita: Awesome, I'll make the changes

jeromecambon (approved):

modules/applications/examples/ui-builder/development/docker-compose.yml
csouchet, line 4: **nitpick:** The `latest` tag doesn't exist.  
The last tag is `1.0.0`.

modules/applications/examples/ui-builder/development/docker-compose.yml
csouchet, line 18: **nitpick:** The `latest` tag doesn't exist.  
The last tag is `1.0.0`.

modules/applications/examples/ui-builder/development/docker-compose.yml
csouchet: After analysing of the UIB promote job, we think it just a missing to promote the Docker image as `latest`.  
So, no need to change the documentation 😊

PR #2897, opened 2024-10-23
chore(uib): update production docker-compose.yml

Shmayro:
As all traffic must be through the uib-proxy, we should not expose ports of uib and bonita.

### DIFF
--- a/modules/applications/examples/ui-builder/production/docker-compose.yml
+++ b/modules/applications/examples/ui-builder/production/docker-compose.yml
@@ -5,13 +5,9 @@ services:
     container_name: bonita-ui-builder
     environment:
       BONITA_DEV_MODE: false
-      BONITA_API_URL: http://host.docker.internal:8080/bonita/API
+      BONITA_API_URL: http://bonita-app:8080/bonita/API
       APPSMITH_ENCRYPTION_PASSWORD: <encryption-password>
       APPSMITH_ENCRYPTION_SALT: <encryption-salt>
-    ports:
-      - "8090:80"
-    extra_hosts:
-      - "host.docker.internal:host-gateway"
     depends_on:
       bonita-app:
         condition: service_healthy
@@ -22,8 +18,11 @@ services:
     ports:
       - "443:443"
       - "80:80"
-    extra_hosts:
-      - "host.docker.internal:host-gateway"
+    environment:
+      - BONITA_HOST=bonita-app
+      - BONITA_PORT=8080
+      - UIB_HOST=bonita-ui-builder
+      - UIB_PORT=80
 
   bonita-app:
     # Replace below with your image name and version, as built following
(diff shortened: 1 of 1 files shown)


marcobonita (approved):

modules/applications/examples/ui-builder/production/docker-compose.yml
```diff
@@ -5,13 +5,9 @@ services:
     container_name: bonita-ui-builder
     environment:
       BONITA_DEV_MODE: false
-      BONITA_API_URL: http://host.docker.internal:8080/bonita/API
+      BONITA_API_URL: http://bonita-app:8080/bonita/API
       APPSMITH_ENCRYPTION_PASSWORD: <encryption-password>
       APPSMITH_ENCRYPTION_SALT: <encryption-salt>
-    ports:
```
csouchet, line 11: **question:** I don't understand why we should remove the expose ports. It will not working ?

modules/applications/examples/ui-builder/production/docker-compose.yml
```diff
@@ -32,8 +31,4 @@ services:
     volumes:
       - <YOUR_LICENSE_FOLDER>:/opt/bonita_lic
     environment:
-      - <ANY_STANDARD_BONITA_ENV_VARIABLE_HERE>
-    ports:
```
csouchet, line 36: **suggestion:** I don't understand why we should remove the expose ports. It will not working ?

csouchet (commented):
I have some questions 😊

modules/applications/examples/ui-builder/production/docker-compose.yml
Shmayro: as ui-builder needs to contact several endpoints to achieve login; Using the port directly will not work correctly ...
The login mechanism will not work if we use directly the UIB port.
So, it is better to expose only the proxy port because all traffic must pass only from it.

modules/applications/examples/ui-builder/production/docker-compose.yml
Shmayro: It's possible to keep it but; Bonita is already accessible via the proxy using /bonita path

modules/applications/examples/ui-builder/production/docker-compose.yml
csouchet: But, the proxy image needs the exposed ports of other image to access to them, no ?

Shmayro: It's no required; as you see i used services names with internal ports
services in the same docker network can communicate using service name and internal ports 😎  🪄 
service name is the hostname by default of the service 😄 
(see env vars introduced in the env vars of the proxy)

csouchet: It's too magic 😂🤩

csouchet (approved):

PR #2900, opened 2024-11-07
docs: faq and guide updates

marcobonita:


modules/applications/pages/ui-builder/how-to-share-data-across-pages.adoc
```diff
@@ -20,20 +20,33 @@ image:ui-builder/guides/configure-query-params.png[configure-query-params]
 The `navigate to` function will redirect to the chosen page. The query params will be passed between the source and the target page's URL: `your-application/target-page/edit?xx=value`
 
 
-== Using JavaScript objects
+== Store data using Store functions
+
+The `storeValue()` function stores the data in the browser’s local storage as key-value pairs that represent storage objects and can be later accessed anywhere in the application.
+
+If you want to store the text of an input widget, you can use storeValue() as shown below:
```
csouchet: **suggestion:** 
```suggestion
If you want to store the text of an input widget, you can use `storeValue()` as shown below:
```

csouchet (approved):

modules/applications/pages/ui-builder/how-to-share-data-across-pages.adoc
```diff
@@ -20,20 +20,33 @@ image:ui-builder/guides/configure-query-params.png[configure-query-params]
...
      }
  }
 ----
 
-Once done, retrieve data using: `{{appsmith.store.keyName}}` in your queries and other widgets.
+Once done, retrieve stored data using `{{appsmith.store.keyName}}` in your queries and other widgets.
 
+Instead of writing code, you can also store data from the widget's settings, see our xref:how-to-control-visibility-of-widgets.adoc#_using_the_storevalue_function[other guide] for more details.
 
+Please note that you can use the `clearStore()` function which clears all the data that was stored in the local storage of the browser, as well as the `removeValue()` function which clears the value associated with a specified key.
```
jeromecambon, line 39: I'm wondering if we should reference the Appsmith documentation at the end of this section.
"For more details, see https://docs.appsmith.com/reference/appsmith-framework/widget-actions/store-value"

jeromecambon: More generally, we could provide in our doc the most common advices or tips, but I think it would be good to reference the Appsmith doc for more complete explanations. And there are also many interesting videos.
I don't think we want to re-write the full Appsmith doc...

marcobonita: You're right, I've made the changed.
However I think we do want to keep a good amount of content in our documentation. From what I've noticed, customers will never think about going to the Appsmith documentation for help (on purpose). This is a Bonitasoft product so they only want to rely on the Bonitasoft documentation

vhemery: IMHO this whole section should be just
See this https://docs.appsmith.com/reference/appsmith-framework/widget-actions/store-value[dedicated guide].

PR #2919, opened 2024-11-29
refactor: move images of the "runtime" module

benjaminParisel:
The images were stored in the Antora legacy folder (assets/images) and included an extra "images" sub-folder.
All images are now in the "images" folder of the module, which is the standard Antora folder.

Covers bonitasoft/bonita-documentation-site#620


tbouffard (approved):
LGTM

PR #2928, opened 2024-12-05
fix(changelog): fix missing Update Tool changelog (2024.3)

educhastenier:
Also documents to new `--update-indexes` option

### DIFF
--- a/modules/version-update/pages/update-tools-changelog.adoc
+++ b/modules/version-update/pages/update-tools-changelog.adoc
@@ -10,6 +10,12 @@ All changes in both tools are logged: from the version that was delivered with t
 This is due to the fact that improvements in any version of the tools can affect all supported Bonita versions. 
 This changelog must be read before xref:update-overview.adoc[updating to a newer version of Bonita].
 
+== 3.6.1 - December 5th, 2024 - Bonita Update Tool changes
+Bonita Update Tool handles updates from all Bonita 7.10 versions up to Bonita 2024.3 (10.2) version.
+
+* [PERFORMANCE] This patch version fixes update versions 9.0 and all of the above, to fix database indexes, in order to improve performance.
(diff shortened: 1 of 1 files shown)


modules/version-update/pages/update-tools-changelog.adoc
```diff
@@ -10,6 +10,12 @@ All changes in both tools are logged: from the version that was delivered with t
 This is due to the fact that improvements in any version of the tools can affect all supported Bonita versions. 
 This changelog must be read before xref:update-overview.adoc[updating to a newer version of Bonita].
 
+== 3.6.1 - December 5th, 2024 - Bonita Update Tool changes
+Bonita Update Tool handles updates from all Bonita 7.10 versions up to Bonita 2024.3 (10.2) version.
+
+* [PERFORMANCE] This patch version fixes update versions 9.0 and all of the above, to fix database indexes, in order to improve performance.
```
ugaston: I would say something like:
_[PERFORMANCE] This patch version fixes database indexes in order to improve performance, when updating to versions 9.0 and above._

modules/version-update/pages/update-tools-changelog.adoc
```diff
@@ -10,6 +10,12 @@ All changes in both tools are logged: from the version that was delivered with t
...
+* [PERFORMANCE] it also introduces a new option `-i` or `--update-indexes` to fix the indexes of the database. This command can be launched without updating Bonita to a newer version. It can also be launched several times without risk: it does nothing if not necessary.
```
ugaston: So, in accordance with my proposition for the first point, I would have this one read:
_[PERFORMANCE] it also introduces a new option `-i` or `--update-indexes` to fix the indexes of the database without actually updating Bonita to a newer version. It can also be launched several times without risk: it does nothing if not necessary._

ugaston (commented):
A bit of rephrasing suggested.
Else, it looks fine to me.

ugaston (approved):
Awesome.
Thanks a lot, Manu!

PR #2943, opened 2024-12-11
chore:  Remove unused images

benjaminParisel:
All deleted images are not used in this version

Covers bonitasoft/bonita-documentation-site#588


tbouffard (approved):
LGTM, the check of Antora references is passing. This confirms that the images are not used.

PR #2944, opened 2024-12-18
refactor: move images of the modules "bcd" and "build-run"

benjaminParisel:
The images were stored in the Antora legacy folder (assets/images) and included an extra "images" sub-folder. All images are now in the "images" folder of the module, which is the standard Antora folder.

All images are used in this version, after running the script bonitasoft/bonita-documentation-site#588, we got:
`[INFO] Total size of unused images: 0 Mo (0 octets)`

Covers bonitasoft/bonita-documentation-site#620


tbouffard (approved):
Nice ❤️

PR #2946, opened 2024-12-19
docs: Autocomplete widget does not update its displayed value

jeromecambon:
Covers [UID-733](https://bonitasoft.atlassian.net/browse/UID-733)

[UID-733]: https://bonitasoft.atlassian.net/browse/UID-733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/pages-and-forms/pages/widgets.adoc
+++ b/modules/pages-and-forms/pages/widgets.adoc
@@ -130,7 +130,7 @@ Autocomplete widget does not update its displayed value
 Autocomplete widget does not update its displayed value when the value is changed by a variable.
 
 [.solution]#Possible solution#
-Make a custom widget from the Autocomplete widget, and Update the controller by replacing the function that set the formatLabel as follows:
+Make a custom widget from the Autocomplete widget, and Update the controller by replacing the function that set the `formatLabel` as follows:
 
 [source,javascript]
 ----
--- a/modules/pages-and-forms/pages/widgets.adoc
+++ b/modules/pages-and-forms/pages/widgets.adoc
@@ -42,6 +42,8 @@ Limitations:
 * The URL should have the file type as extension (e.g. \https://myhost/doc/sample.pdf)
 ====
 
+See [#Troubleshooting UI Designer widgets] to solve common issues with this widget.
+
 [#rich-text-area]
 
 == Rich Text Area widget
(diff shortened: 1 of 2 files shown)


benjaminParisel:
image

I am not satisfied with the result. I suggest to move the troubleshooting section in the dedicated widget part.

I push a commit to do it, feel free to rollback if you think it's better before.

jeromecambon:
@benjaminParisel 
I think that including the troubleshooting in the widget description make the page less readable, and would "pollute" the widget descriptions. However, adding a pointer from the widget description to the dedicated troubleshooting (if any) would provide a good alternative, and avoid to miss it. What do you think?

benjaminParisel:
> @benjaminParisel I think that including the troubleshooting in the widget description make the page less readable, and would "pollute" the widget descriptions. However, adding a pointer from the widget description to the dedicated troubleshooting (if any) would provide a good alternative, and avoid to miss it. What do you think?

Do what you think is better and more readable but please, don't put three troubleshooting titles and icon (like we can see on the screen)

jeromecambon:
> > @benjaminParisel I think that including the troubleshooting in the widget description make the page less readable, and would "pollute" the widget descriptions. However, adding a pointer from the widget description to the dedicated troubleshooting (if any) would provide a good alternative, and avoid to miss it. What do you think?
> 
> Do what you think is better and more readable but please, don't put three troubleshooting titles and icon (like we can see on the screen)

I kept the same formatting already in place, but yes we should use the example we provided here: https://bonitasoft.github.io/bonita-documentation-theme/troubleshooting.html

modules/pages-and-forms/pages/widgets.adoc
```diff
@@ -42,6 +42,8 @@ Limitations:
 * The URL should have the file type as extension (e.g. \https://myhost/doc/sample.pdf)
 ====
 
+See [#Troubleshooting UI Designer widgets] to solve common issues with this widget.
```
benjaminParisel:
```suggestion
See <<troubleshooting-ui-designer-widgets>> to solve common issues with this widget.
```

Example could be found here, https://bonitasoft.github.io/bonita-documentation-theme/links.html#_anchors_to_navigate_inside_the_same_document

benjaminParisel (changes requested):

benjaminParisel (approved):

PR #2952, opened 2025-01-07
feat: How to stream runtime logs as JSON ouptut to a remote server

benjaminParisel:
* Add a section on logging page to explain how to configure log4j to stream the log as JSON output to a remote server

### DIFF
--- a/modules/setup-dev-environment/pages/logging.adoc
+++ b/modules/setup-dev-environment/pages/logging.adoc
@@ -46,6 +46,32 @@ Here is an example of the Log4j2 configuration for the logger used in the exampl
 <Logger name="org.bonitasoft.groovy.script" level="TRACE"/>
 ----
 
+
+==== How to configure the JSON output
+
+To configure a JSON appender, you need to add a **File** appender with a JSONLayout to your log4j2-appenders.xml file in `Appenders` section. Here is an example configuration:
+
+[source, xml]
+----
+<File name="JsonFileAppender" fileName="${baseDir}/bonita.json">
+    <JSONLayout compact="true" eventEol="true" properties="true" stacktraceAsString="true">
+        <KeyValuePair key="timestamp" value="$${date:yyyy-MM-dd'T'HH:mm:ss.SSSZ}"/>
+    </JSONLayout>
+</File>
+----
+
+Then you need to add in `Loogers` block
+[source, xml]
+----
+<Loggers>
+        <Root level="INFO">
+            <!-- [...] -->
+            <AppenderRef ref="JsonFileAppender"/> <!-- Add this AppenderRef -->
+            <!-- [...] -->
+        </Root>
+    </Loggers>
+----
+
 === Bonita Studio
 
 To change the Bonita Studio log level (usually to debug Studio), go to the Studio installation folder, configuration directory, and edit the `config.ini` file. You need to set the property `eclipse.log.level` to one of the following values: INFO, WARNING or ERROR.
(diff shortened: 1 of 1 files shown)


modules/setup-dev-environment/pages/logging.adoc
```diff
@@ -46,6 +46,35 @@ Here is an example of the Log4j2 configuration for the logger used in the exampl
 <Logger name="org.bonitasoft.groovy.script" level="TRACE"/>
 ----
 
+
+==== Stream logs to a remote server
+
+You need to declare a socket in `log4j-appenders.xml` with a `JSONLayout` to stream to a remote server.
```
rbioteau: Note that the SocketAppender and the JSONLayout are 2 distinct concept. You may use Socket with plain text logs.

vhemery: I would split in 2 distinct paragraphs. One about how to use the JSONLayout (which we already have in default configuration but is not used in Loggers). Another about the Socket for streaming.

Alternatively, we can also just update the log4j2-appenders.xml to add the Socket example appender. Make the appenders comments more self-explanatory.
And document here that we can add Loggers to use the appender we want.

modules/setup-dev-environment/pages/logging.adoc
```diff
@@ -46,6 +46,35 @@ Here is an example of the Log4j2 configuration for the logger used in the exampl
...
+----
+<Loggers>
+        <Root level="INFO">
+            <!-- [...] -->
+            <AppenderRef ref="runtimeLogSocket"/> <!-- Add this AppenderRef -->
+            <!-- [...] -->
+        </Root>
+    </Loggers>
+----
+
```
vhemery:
```suggestion

==== Use JSON format for logs

The default configuration contains an unused `Console-JSON` appender which formats logs as JSON.

[source, xml]
----
    <Appenders>
        <!-- use json logger if json output is required -->
        <Console name="Console-JSON" target="SYSTEM_OUT">
            <JSONLayout compact="true" eventEol="true" properties="true" stacktraceAsString="true">
                <KeyValuePair key="timestamp" value="$${date:yyyy-MM-dd'T'HH:mm:ss.SSSZ}"/>
            </JSONLayout>
        </Console>
        <!-- [...] -->
    </Appenders>
----

To use it, you can update the `AppenderRef` in the `Loggers` block or add a new one.

[source, xml]
----
    <Loggers>
        <Root level="INFO">
            <AppenderRef ref="Console-JSON"/>
        </Root>
    </Loggers>
----

==== Stream logs to a remote server

You need to declare a Socket appender in `log4j-appenders.xml` to stream to a remote server. You can choose to use Socket with various formats, such as JSONLayout or PatternLayout.

As example, this configuration is used to stream JSON-formatted log entries to a remote server, such as Logstash, for further processing and analysis.

[source, xml]
----
<Socket name="runtimeLogSocket" host="<host>" port="<port>">
    <JSONLayout compact="true" eventEol="true" properties="true" stacktraceAsString="true">
        <KeyValuePair key="timestamp" value="$${date:yyyy-MM-dd'T'HH:mm:ss.SSSZ}"/>
    </JSONLayout>
</Socket>
----

Then, do not forget to reference this new appender in `Loggers` block:

[source, xml]
----
    <Loggers>
        <Root level="INFO">
            <!-- [...] -->
            <AppenderRef ref="runtimeLogSocket"/> <!-- Add this AppenderRef -->
            <!-- [...] -->
        </Root>
    </Loggers>
----

```

vhemery (changes requested):
Suggestion with separation of JSON & Socket concepts.

vhemery (approved):

PR #2986, opened 2025-02-24
feat(bcd): delete modules/bcd/pages/builder.adoc

Yves-Jacq:
Remove the page "builde.adoc" because LA builder is no longer supported since 2023.2


uguy:
Replaced by https://github.com/bonitasoft/bonita-doc/pull/2989

PR #2992, opened 2025-03-06
fix(software): fix wrong CentOS version

educhastenier:
that was probably retrograded erroneously.

This commit restores the version supported from 7 to 8.4


abirembaut (approved):

vhemery (approved):
LGTM

PR #3006, opened 2025-03-28
docs(perf tunning): fix performance tuning configuration doc

educhastenier:
We missed to removed an obsolete configuration.

Also, improve scheduler tuning configuration documentation.


JeremJR (approved):
LGTM

PR #3012, opened 2025-04-04
feat: improve how the `la-deployer` version is documented

tbouffard:
Compatibility matrix: only reference the la-deployer 1.1.x which can manage all supported versions of Bonita (and some older versions).

In the example demonstrating the usage of the deployer, use a specific version to ensure that the documented command can be copied and used as it successfully.

The version is stored in an AsciiDoc attribute to ease maintenance: it can be used in various pages and updated easily in the next versions of the documentation.

### Notes

Closes https://bonitasoft.atlassian.net/browse/RUNTIME-1945

To be reviewed by the Support Team as they requested improvements of the related pages.

> [!IMPORTANT]
> When merging this in next versions of Bonita, the AsciiDoc attribute may not be applied to the antora.yml. In this case, the attribute resolution will fail and the new site won't be deployed.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -8,6 +8,7 @@ asciidoc:
     bonitasoft-docker-repository: '{bonitasoft-registry}/docker'
     bonitaTechnicalVersion: '9.0.0' # latest public community version
     javadocVersion: '9.0'
+    laDeployerVersion: '1.1.0'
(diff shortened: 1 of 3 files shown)


JJargotExternalDanone:
Hi Thomas,
This looks awesome !

educhastenier (approved):

PR #3033, opened 2025-05-15
Adding section on init script Update bonita-docker-installation.adoc

jjargot:
Adding a section related to the init scripts that help customers perform any critical operations before your services are fully up and running


jjargot:
I would like to fix the title but there is no button / link that I could find to do this

jjargot:
wrong title
need to show associated volumes to used with docker compose

PR #3059, opened 2025-07-08
feat: prepare release 2024.1-u7/10.0.7

abirembaut:


modules/ROOT/pages/release-notes.adoc
```diff
@@ -119,6 +119,10 @@ The property allowing to enable/disable xref:identity:rest-api-authorization.ado
 
 == Bug fixes
 
+=== Fixes in Bonita 2024.1-u7 (2025-07-07)
```
tbouffard, line 122: **nitpick:** update the date when the binaries are available to customers

tbouffard (changes requested):

modules/ROOT/pages/release-notes.adoc
abirembaut: I have always used the date of the build actually. It avoid confusion if the release is not published right away  for any reason and there are several versions publicly release at the same time

tbouffard: OK, as you wish
For other maintenance versions, I used the date when the binaries were publicly available. This may be something we should mention in our internal process.

tbouffard (approved):
LGTM

PR #3104, opened 2025-08-29
ci(2025.2): trigger publish to production when contents change

tbouffard:
Use generic glob for branch pattern and github context instead of using hard coded branch names.
In the future, this will prevent from having to update the workflow definition.

### Notes

Covers https://github.com/bonitasoft/bonita-documentation-site/issues/472


akantcheff (approved):
Perfectly what I want with the BTT doc as well :+1:

PR #3164, opened 2025-11-21
feat(oracle): add support for Oracle 23-ai

educhastenier:


modules/runtime/pages/hardware-and-software-requirements.adoc
```diff
@@ -83,7 +83,7 @@ Software required for Bonita Platform.
 | 2022
 
 | Oracle
-| 19c (19.3.0.0.0) (see note 5)
+| 19c (19.3.0.0.0) & 23-ai (23.9.0.0) (see note 5)
```
akantcheff, line 86: Should we complete the release notes as well?

educhastenier: I don't think so because for me, the release notes are mainly for the first update -u0.
But we can discuss it

abirembaut (approved):

PR #3256, opened 2026-03-05
feat: Externalise as an independent component the UIB documentation

benjaminParisel:
Prepare externalization of bonita-ui-builder documentation


tbouffard (approved):
LGTM
✔️ 1st step of the externalization plan

PR #3273, opened 2026-03-24
docs: add proxy-url property to OIDC configuration reference

cvidaillac:
Add documentation for the proxy-url property in keycloak-oidc.json, which configures an HTTP proxy for outbound connections to the OIDC identity provider.

This property is required in hardened network environments where the Bonita server cannot reach the OIDC provider directly. Standard JVM proxy properties (-Dhttp.proxyHost, -Dhttps.proxyHost) set via CATALINA_OPTS do not apply, as the OIDC adapter manages its own HTTP client.

Applies to versions: 2023.2, 2024.1, 2024.2, 2024.3, 2025.1, 2025.2

Ref: Customer feedback from support case #392035

### DIFF
--- a/modules/identity/pages/single-sign-on-with-oidc.adoc
+++ b/modules/identity/pages/single-sign-on-with-oidc.adoc
@@ -235,6 +235,9 @@ NOTE: Activating this option means any user authorized by the IdP to access Boni
     "^/bonita/(.*)$" : "/$1"
   }
 ----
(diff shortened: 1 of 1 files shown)


ugaston (approved):
Looks good to me.

Good catch.

abirembaut (changes requested):
Good catch !
The contents looks good to me but the Note appears as a code snippet. it would be better to display it as text rather than code:

Image

cvidaillac (commented):
Approve

abirembaut (approved):

PR #3300, opened 2026-04-13
docs: document how to backfill tracking for legacy BDM data — BPA-441

tmarie-bonitasoft:
## Summary

Documents the manual migration procedure to bring BDM instances that existed *before* the data retention feature under retention control.

Added to the *BDM data retention* page (`#legacy-data` anchor):

- Explanation of why pre-existing BDM instances are not tracked.
- IMPORTANT admonition making the two-database split explicit: the BDM tables live in the *Business Data database* while the `data_retention_bdm_tracking` and `sequence` tables live in the *Bonita engine database*, so the backfill is split into two scripts run on two distinct connections (no cross-database joins).
- Three-step procedure with runnable PostgreSQL examples:
  - **Step 1** — export legacy persistence IDs and reference dates from the Business Data database to a CSV (`psql \copy ... TO ...`).
  - **Step 2** — load the CSV into a staging table on the Bonita engine database, reserve N IDs from the `sequence` row (`id = 9`), and insert N tracking rows with IDs in that reserved range.
  - **Step 3** — verification split into two count queries, each labeled with the database it must run against.
- Per-engine reference tables for both bulk-export and bulk-load tools (PostgreSQL, MySQL, Oracle, SQL Server), plus the SQL Server equivalent for `SELECT ... FOR UPDATE` (`WITH (UPDLOCK, HOLDLOCK)`).

Also fixes two minor accuracy issues: the sequence range-size default is `100` (configurable via `bonita.platform.sequence.defaultRangeSize`), not `20`; and the CAUTION block now asks for a backup of *both* databases.

Cross-linked from the GDPR design guide.

Closes [BPA-441](https://bonitasoft.atlassian.net/browse/BPA-441).

## Test plan
- [x] Built locally with Antora and reviewed in the browser
- [ ] SQL example validated against a legacy Bonita deployment (reviewer step)

[BPA-441]: https://bonitasoft.atlassian.net/browse/BPA-441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/runtime/pages/bdm-data-retention.adoc
+++ b/modules/runtime/pages/bdm-data-retention.adoc
@@ -55,6 +55,103 @@ Each cleanup run produces one xref:api:queriable-logging.adoc[queriable log] ent
 
 These entries can be queried through the existing `/API/system/log` REST endpoint and are displayed in the *Audit log* tab of the *Data Retention (GDPR)* administration page.
 
+[#legacy-data]
+== Handling legacy data
+
+BDM instances that were created *before* the tracking mechanism was introduced do not have a corresponding row in the Bonita table `data_retention_bdm_tracking`. As a consequence, retention rules configured on their object type cannot identify them as expired, and they are never deleted by the cleanup job.
+
+To bring legacy data under retention control, populate the tracking table manually with one row per legacy instance. The reference dates (`created_at` and `last_modified_at`) must be provided explicitly — use the best approximation available in your existing data (for example, a business creation date column, or a fallback value acceptable to your compliance policy).
+
+[CAUTION]
+====
+*Stop the Bonita platform* before running the backfill, and make a full backup of the database first. Bonita manages IDs for `data_retention_bdm_tracking` through its internal `sequence` table, and caches the next available range in memory; a manual backfill while the platform is running would conflict with the cached range and may create duplicate IDs.
+====
+
+=== How Bonita generates IDs for the tracking table
+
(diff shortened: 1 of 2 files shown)


educhastenier (commented):
explanations are correct, even though the entire process seems complicated to me: I hope our customers are techy enough

modules/runtime/pages/bdm-data-retention.adoc
```diff
@@ -55,6 +55,103 @@ Each cleanup run produces one xref:api:queriable-logging.adoc[queriable log] ent
...
+
+-- Lock the sequence row and read the next available ID (e.g. 101)
+SELECT nextid FROM sequence WHERE id = 9 FOR UPDATE;
+
+-- Reserve one ID per legacy Invoice that is not yet tracked (e.g. 500 rows => reserve 500 IDs)
+UPDATE sequence
+   SET nextid = nextid + (
+        SELECT COUNT(*) FROM invoice i
+         WHERE NOT EXISTS (
+            SELECT 1 FROM data_retention_bdm_tracking t
```
akantcheff: [issue](https://pullpo.io/cc?l=issue) 
`data_retention_bdm_tracking` and `invoice` tables are not in the same schema... That won't work...

akantcheff (changes requested):
[issue(blocking)](https://pullpo.io/cc?l=issue&d=blocking)

abirembaut (approved):

PR #3344, opened 2026-05-07
feat: add Bonita 2021.1, 2021.2, 2022.1, 2022.2 to archives

tbouffard:


tbouffard:
> [!NOTE]
> https://github.com/bonitasoft/bonita-doc/pull/3344#issuecomment-4395352025 can be ignored. We must use Bonita BPM name because this was the name of the product at the time the related versions have been released

benjaminParisel (approved):

PR #3351, opened 2026-05-13
feat: add central 1.0 to archives

tbouffard:
The central archives are stored in a dedicated page.
The bonita archives also live in a dedicated page, and a new home page references all archive pages.

### Notes

The Bonita archive page includes the "Bonita BPM" wording which is not allowed. We have no choice here, because it is the former name of the Bonita product and we must use it in the page. 
So, this Contribution Guidelines checks error can be ignored.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -2,7 +2,7 @@ name: bonita
 title: Bonita
 version: 0
 display_version: Archives
-start_page: ROOT:archives.adoc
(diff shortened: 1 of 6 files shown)


benjaminParisel (approved):

benjaminParisel:
The failed check is ok and under control, it's because it's the oldest product name.

PR #3425, opened 2026-09-02
docs(runtime): document the Hazelcast 5.7.0 update and the cluster work recovery properties

abirembaut:
**On hold until bonitasoft/bonita-engine-sp#4108, bonitasoft/bonita-engine-sp#4112 and bonitasoft/bonita-engine-sp#4113 are merged into `10.2.x`.** The work-recovery paragraphs and the "Work recovery is more robust" release-note bullet describe the first two; the Hazelcast interface properties in the hardening paragraph and the "Security hardening of the Hazelcast member" bullet describe the third. Merging this before them would document behaviour 2024.3-u10 does not have.

Documents the Hazelcast update shipping in **2024.3-u10** (technical id 10.2.10).

## Release notes (`modules/ROOT/pages/release-notes.adoc`)

A `2024.3-u10` section recording the Hazelcast update from 5.4.0 to 5.7.0, plus an **Updating a cluster to 2024.3-u10** block covering what an administrator has to act on:

- **all nodes must be stopped** to update, because Hazelcast does not support updating cluster members one at a time, so a 5.4.0 node cannot join a 5.7.0 cluster;
- **no configuration change is required** - the two new properties ship with defaults;
- a **harmless schema warning** at every node start, because the stored configuration still names the previous schema and the update does not rewrite it;
- **two diagnostics change**: `Number of locks` in the Hazelcast statistics file changes unit and scope, and an expired lock lease is now reported as an error rather than a warning.

Scoped to Subscription editions with the Cluster feature; single-node installations are unaffected and Community does not use Hazelcast.

## Cluster page (`modules/runtime/pages/install-a-bonita-bpm-cluster.adoc`)

A new **Work recovery when a node leaves the cluster** section under *Cluster management*, documenting the two new properties with their defaults and when to change them:

- `bonita.platform.cluster.work.recoveryLockTimeoutSeconds` (30)
- `bonita.platform.cluster.work.recoveryLockLeaseSeconds` (14400)

The work-recovery section also describes, from bonitasoft/bonita-engine-sp#4112, that the recovery runs in the background, that work it leaves behind is tried again at the next departure or start, and that a node resuming after a tenant pause resubmits the work that arrived meanwhile; the release notes carry one bullet for it.

The sizing advice for the lease reflects bonitasoft/bonita-engine-sp#4108: the handover of a departed node's work stops by itself at 90% of the lease, so the lease is a working limit rather than a backstop, and the page says to keep it in hours rather than minutes.

It also documents `bonita.platform.cluster.lock.leaseTimeSeconds` (600) beside them - that property already existed and appeared nowhere in the documentation.

## Version correspondence table (`modules/version-update/pages/product-versioning.adoc`)

The `2024.3-u10` / `10.2.10` row.

## Notes for the reviewer

- **The release date is `TBD`** in the release-notes heading and in the correspondence table, and needs filling in when the release is dated.
- Kept deliberately non-technical: the section explains what the properties bound and when to raise them, not how the locking works.

### DIFF
--- a/modules/ROOT/pages/release-notes.adoc
+++ b/modules/ROOT/pages/release-notes.adoc
@@ -82,6 +82,48 @@ Note that this new API is now also available in the Community edition and provid
 
 == Bug fixes
 
+=== Fixes in Bonita 2024.3-u10 (TBD)
+
+* Hazelcast, the library used by the Cluster feature, is updated from 5.4.0 to 5.7.0. If you run a cluster, read <<v2024-3-u10-cluster-update,Updating a cluster to 2024.3-u10>> before updating.
+* Security hardening of the Hazelcast member. On a single node, the member Bonita starts for its second-level cache now
+listens on loopback only instead of on every interface.
+Set `bonita.platform.hazelcast.interfaces.list` to an empty value in `bonita-platform-sp-custom.properties` to restore
+the previous behaviour, for a deployment that attaches a Hazelcast client on purpose. In a cluster,
+`bonita.platform.cluster.hazelcast.interfaces.list` now binds the member to one address of the node matching the
+patterns it lists, which it never did before; its default no longer names an interface. If your cluster configuration sets that property, verify its
+value or unset it before updating: a value that matches no interface of the node, harmless until now, stops the node
+from starting. See
+xref:runtime:install-a-bonita-bpm-cluster.adoc#hazelcast-port-hardening[Allow communication between servers] for how to
+restrict the port.
+
+[#v2024-3-u10-cluster-update]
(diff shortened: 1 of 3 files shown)


abirembaut:
Added the `2024.3-u10` / `10.2.10` row to `modules/version-update/pages/product-versioning.adoc`.

Two values in that row still need confirming when the release is assembled:

- **Release date** is `TBD`, as in the release-notes section heading.
- **UI Designer version** is carried forward as `1.19.1`. It has been 1.19.1 since 2024.3-u5, so this is the likely value rather than a known one.

akantcheff (approved):

